PR #3, opened 2019-07-06
clean up and add tests

Tianhao-Gu:


Tianhao-Gu:
@bio-boris do you mind taking a look? It should have everything up to submit condor job.

lib/execution_engine2/utils/MongoUtil.py
```diff
@@ -0,0 +1,158 @@
...
+        self.mongo_database = config['mongo-database']
+        self.mongo_user = config['mongo-user']
+        self.mongo_pass = config['mongo-password']
+        self.mongo_authmechanism = config['mongo-authmechanism']
+
+        self.mongo_collection = config['mongo-collection']
+
+        try:
+            start_local = int(config.get('start-local-mongo', 1))
+        except Exception:
```
bio-boris: I think we should throw the exception here

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -0,0 +1,116 @@
...
+        inputs.wsid = job.wsid
+        inputs.method = params.get('method')
+        inputs.params = params
+        inputs.service_ver = params.get('service_ver')
+        inputs.app_id = params.get('app_id')
+
+        job.job_input = inputs
+        job.job_output = output
+
+        insert_rec = self.mongo_util.insert_one(job.to_mongo())
```
bio-boris: Is there a reason you want to insert one rather than just  do a Job.save()?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -0,0 +1,116 @@
...
+        self._check_ws_ojects(params.get('source_ws_objects'))
+
+        # update service_ver
+        git_commit_hash = self._get_module_git_commit(method, params.get('service_ver'))
+        params['service_ver'] = git_commit_hash
+
+        # insert initial job document
+        job_id = self._init_job_rec(user_id, params)
+
+        print(client_groups)
```
bio-boris: print can be removed

bio-boris:
@Tianhao-Gu Do you mind making the tests able to run without kb-sdk? I would like to be able to run these tests from my IDE and step through them with a debugger.

lib/execution_engine2/utils/MongoUtil.py
Tianhao-Gu: 👍

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

PR #7, opened 2019-07-19
CondorUtils

bio-boris:


deploy.cfg
```diff
@@ -8,21 +9,47 @@ srv-wiz-url = {{ srv_wiz_url }}
 njsw-url = {{ njsw_url }}
 auth-service-url = {{ auth_service_url }}
 auth-service-url-allow-insecure = {{ auth_service_url_allow_insecure }}
+scratch = /kb/module/work/tmp
 
-{% if catalog_url %}
-catalog-url = {{ catalog_url }}
```
Tianhao-Gu: Don't we need catalog and mongo config here?

bio-boris: Yea I plan to fix this configs for deployment soon. It's very likely they will change again today.

PR #15, opened 2019-07-30
Authorization and log models

bio-boris:


bio-boris:
* I broke SDK and JobRunner tests, if you can fix them that would be great
* Can we squash and merge before we merge this?

Tianhao-Gu:
@bio-boris I fixed SDKMethodRunner and MongoUtil tests.

PR #16, opened 2019-07-30
using mongo engine to save

Tianhao-Gu:


lib/execution_engine2/utils/MongoUtil.py
```diff
@@ -60,10 +61,23 @@ def _get_collection(
                 authSource=mongo_database,
                 authMechanism=mongo_authmechanism,
             )
+
+            connect(
```
bio-boris: How do we disconnect this connection?

Tianhao-Gu:
@bio-boris do you mind taking an another look?

bio-boris:
Very interesting! I look forward to trying this context manager.

PR #19, opened 2019-08-05
Adding job logs and auth

bio-boris:


bio-boris:
@Tianhao-Gu Could you please review?
Also could you please 1) use count_documents({}) instead of count(), as count is deprecated
2) Don't rely on count_documents, save it first, as it may change based on other tests and the test becomes flaky
3) Could you please test the precommit hook by installing it in precommit.sh, and then when you do a commit -m it will tell you if your files failed
4) I might have screwed up some tests during a merge, if you spot that please let me know

bio-boris:
Never mind about #2, I see you started doing that now.

PR #20, opened 2019-08-07
add get_job_params endpoint

Tianhao-Gu:


Tianhao-Gu:
@bio-boris mind taking a look?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -261,3 +264,45 @@ def run_job(self, params, ctx):
...
+
+            job_params["method"] = job_input.method
+            job_params["params"] = job_input.params
+            job_params["service_ver"] = job_input.service_ver
+            job_params["app_id"] = job_input.app_id
+            job_params["wsid"] = job_input.wsid
+            job_params["parent_job_id"] = job_input.parent_job_id
+            job_params["source_ws_objects"] = job_input.source_ws_objects
+
+        return job_params, job_config
```
bio-boris: job_config unused?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -261,3 +264,45 @@ def run_job(self, params, ctx):
...
+
+        with self.get_mongo_util().me_collection(self.config["mongo-jobs-collection"]):
+
+            try:
+                job = Job.objects(id=job_id)[0]
+            except Exception:
+                raise ValueError("Unable to find job:\nError:\n{}".format(traceback.format_exc()))
+
+            job.status = status
+            job.updated = datetime.utcnow()
```
bio-boris: Maybe we should edit the model so we don't have to do job.updated manually each time

https://stackoverflow.com/questions/12230022/update-datetime-field-in-mongoengine-with-auto-now-add

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: it's not used. I removed it from spec.

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍 sounds like a good idea. Implemented auto update upon saving.

PR #21, opened 2019-08-09
add get_job_status and remove unnecessary funs from spec

Tianhao-Gu:


Tianhao-Gu:
@bio-boris removed `update_job, finish_job, cancel_job, check_job_canceled` endpoint and added `get_job_status` endpoint.

bio-boris:
We need finish job to update the finish time
We need cancel job to Send a signal to the scheduler to cancel the job

test/SDKMethodRunner_test.py
```diff
@@ -143,96 +141,94 @@ def test_get_module_git_commit(self):
...
 
     def test_update_job_status(self):
 
-        self.assertEqual(self.test_collection.count(), 0)
-        job_id = self.create_job_rec()
-        self.assertEqual(self.test_collection.count(), 1)
+        with self.mongo_util.me_collection(self.cfg["mongo-jobs-collection"]):
 
-        runner = self.getRunner()
+            ori_job_count = Job.objects.count()
```
bio-boris: can you please use countDocuments, count is deprecated

bio-boris:
Looks like there is a conflict now

test/SDKMethodRunner_test.py
```diff
@@ -378,55 +358,48 @@ def test_run_job_and_add_log(self, condor_mock):
...
-        self.assertEqual(params["wsid"], self.ws_id)
-        self.assertEqual(params["method"], "MEGAHIT.run_megahit")
-        self.assertEqual(params["app_id"], "MEGAHIT/run_megahit")
-        self.assertEqual(params["service_ver"], "2.2.1")
-        self.assertCountEqual(params["source_ws_objects"], ["a/b/c", "e/d"])
-        self.assertEqual(params["parent_job_id"], "9998")
-
-        self.test_collection.delete_one({"_id": ObjectId(job_id)})
-        self.assertEqual(self.test_collection.count_documents({}), count)
+        with self.mongo_util.me_collection(self.cfg["mongo-jobs-collection"]):
```
bio-boris: You could use `self.mongo_util.mongo_engine_connection` here

test/SDKMethodRunner_test.py
```diff
@@ -378,55 +358,48 @@ def test_run_job_and_add_log(self, condor_mock):
...
+                                    "source_ws_objects", "parent_job_id"]
+            self.assertCountEqual(params.keys(), expected_params_keys)
+            self.assertEqual(params["wsid"], self.ws_id)
+            self.assertEqual(params["method"], "MEGAHIT.run_megahit")
+            self.assertEqual(params["app_id"], "MEGAHIT/run_megahit")
+            self.assertEqual(params["service_ver"], "2.2.1")
+            self.assertCountEqual(params["source_ws_objects"], ["a/b/c", "e/d"])
+            self.assertEqual(params["parent_job_id"], "9998")
+
+            Job.objects.get(id=job_id).delete()
```
bio-boris: You could use `get_job_log` here

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -467,3 +467,20 @@ def update_job_status(self, job_id, status):
...
+    def get_job_status(self, job_id):
+        returnVal = dict()
+
+        if not job_id:
+            raise ValueError("Please provide valid job_id")
+
+        with self.get_mongo_util().me_collection(self.config["mongo-jobs-collection"]):
+
+            try:
+                job = Job.objects(id=job_id)[0]
```
bio-boris: I think its better to use  `self.get_mongo_util().me_collection.get_job` to keep database stuff in that utils file

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -454,16 +452,32 @@ def update_job_status(self, job_id, status):
...
-                job = Job.objects(id=job_id)[0]
-            except Exception:
-                raise ValueError(
-                    "Unable to find job:\nError:\n{}".format(traceback.format_exc())
-                )
+        try:
+            job = self.get_mongo_util().get_job(job_id=job_id)
+        except Exception:
+            raise ValueError(
+                "Unable to find job:\nError:\n{}".format(traceback.format_exc())
```
bio-boris: I think it returns a None if it doesn't find a job.

Tianhao-Gu: @bio-boris yea. But only when you supply valid ObjectID. Otherwise,`mongoengine.errors.ValidationError: 'aaaa' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string`

bio-boris: Good point. You might want to throw an error if the job is None though.

test/SDKMethodRunner_test.py
Tianhao-Gu: hmm I found it in latest mongoengine doc. Besides Job.objects.count_documents() is not defined. 

http://docs.mongoengine.org/guide/querying.html#counting-results

bio-boris: For pymongo it is deprecated.

https://docs.mongodb.com/manual/reference/method/db.collection.count/

bio-boris: You are right, I was not paying enough attention my comment was wrong.

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

test/SDKMethodRunner_test.py
Tianhao-Gu: 👍

test/SDKMethodRunner_test.py
Tianhao-Gu: 👍

Tianhao-Gu:
@bio-boris Do you mind taking another look? I added finish and start_job.

lib/execution_engine2/execution_engine2Impl.py
```diff
@@ -301,50 +292,52 @@ def get_job_logs(self, ctx, params):
...
+           optional if job is finished with error) -> structure: parameter
+           "job_id" of type "job_id" (A job id.), parameter "error_message"
+           of String
         """
         # ctx is the context object
-        # BEGIN finish_job
-        # END finish_job
+        #BEGIN finish_job
+        mr = SDKMethodRunner(self.config)
+        mr.finish_job(params.get('job_id'), error_message=params.get('error_message'))
```
bio-boris: This should default to empty string or None if not supplied

lib/execution_engine2/utils/MongoUtil.py
```diff
@@ -146,13 +146,27 @@ def get_job_log(self, job_id=None) -> JobLog:
         if job_id is None:
             raise ValueError("Please provide a job id")
         with self.mongo_engine_connection():
-            return JobLog.objects.with_id(job_id)
+            try:
+                job_log = JobLog.objects.with_id(job_id)
```
bio-boris: You could use the get_job_log function function here

lib/execution_engine2/utils/MongoUtil.py
```diff
@@ -146,13 +146,27 @@ def get_job_log(self, job_id=None) -> JobLog:
...
     def get_job(self, job_id=None) -> Job:
         if job_id is None:
             raise ValueError("Please provide a job id")
         with self.mongo_engine_connection():
-            return Job.objects.with_id(job_id)
+            try:
+                job = Job.objects.with_id(job_id)
+            except Exception:
+                raise ValueError(
+                    "Unable to find job:\nError:\n{}".format(traceback.format_exc())
```
bio-boris: If you think we should raise exceptions here for being unable to find a job due to supplying a bad objectid, then we should also raise an exception if the job is None in case the objectID supplied is OK, but that job doesn't exist

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -294,7 +294,14 @@ def check_job_canceled(self, job_id, ctx):
         Check to see if job is terminated by the user
         :return: job_id, whether or not job is canceled, and whether or not job is finished
         """
-        job_status = self.get_mongo_util().get_job(job_id=job_id).status
+
+        job = self.get_mongo_util().get_job(job_id=job_id)
+
+        if not job:
```
bio-boris: Maybe we should just throw the exception in the get_job function if the job is not None so we don't have to check if its None each time.

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -454,16 +457,65 @@ def update_job_status(self, job_id, status):
...
+
+        job = self.get_mongo_util().get_job(job_id=job_id)
+
+        if error_message:
+            job.errormsg = error_message
+            self.get_mongo_util().update_job_status(job_id=job_id, status=Status.error.value)
+        else:
+            self.get_mongo_util().update_job_status(job_id=job_id, status=Status.finished.value)
+
+        job.finished = datetime.utcnow()
```
bio-boris: I think you need the mongo-engine context to save the job object right?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -454,16 +457,65 @@ def update_job_status(self, job_id, status):
...
+            job.running = datetime.utcnow()
+            self.get_mongo_util().update_job_status(job_id=job_id, status=Status.running.value)
+        elif job_status in [Status.created.value, Status.queued.value]:
+            # set job to estimating status
+            job.estimating = datetime.utcnow()
+            self.get_mongo_util().update_job_status(job_id=job_id, status=Status.estimating.value)
+        else:
+            raise ValueError("Unexpected job status: {}".format(job_status))
+
+        job.save()
```
bio-boris: I think you need the mongo-engine context to save the job object right?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -454,16 +457,65 @@ def update_job_status(self, job_id, status):
...
+            raise ValueError("Please provide valid job_id")
+
+        job = self.get_mongo_util().get_job(job_id=job_id)
+        job_status = job.status
+
+        if job_status == Status.estimating.value or skip_estimation:
+            # set job to running status
+            job.running = datetime.utcnow()
+            self.get_mongo_util().update_job_status(job_id=job_id, status=Status.running.value)
+        elif job_status in [Status.created.value, Status.queued.value]:
```
bio-boris: Very good. I think we should first check the elif condition first, since that would follow the logic structure of

runjob -> start_job -> start_job -> finish_job
Created -> Estimating -> Running -> Finished

I think it makes the code a little easier to follow to do it that way. We should probably document the flow of this in the comments for this function too

bio-boris:
Looks good. Needs to add authentication to the endpoints and some feedback given.  You can mock out the auth during unit tests. Also we should add some docstrings for the public methods.

lib/execution_engine2/execution_engine2Impl.py
Tianhao-Gu: sure. But params.get defaults to `None` if `error_message` is not found.

lib/execution_engine2/utils/MongoUtil.py
Tianhao-Gu: 👍  good point.

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍

lib/execution_engine2/execution_engine2Impl.py
bio-boris: You are right, my mistake.

lib/execution_engine2/utils/SDKMethodRunner.py
bio-boris: Sorry this is a typo, "if the job IS none" we should throw an exception

lib/execution_engine2/utils/MongoUtil.py
bio-boris: Sorry this is a typo, "if the job IS none" we should throw an exception

PR #24, opened 2019-08-13
Testing for timestamps fixing flake8

bio-boris:


bio-boris:
@Tianhao-Gu I accidentally merged this one in, can you check it out

PR #25, opened 2019-08-14
add catalog log_exec_stats

Tianhao-Gu:


lib/execution_engine2/models/models.py
```diff
@@ -126,7 +126,8 @@ class Job(Document):
     scheduler_type = StringField()
     scheduler_id = StringField()
     job_input = EmbeddedDocumentField(JobInput, required=True)
-    job_output = EmbeddedDocumentField(JobOutput)
```
Tianhao-Gu: I have zero clue what's going on here. Just like `LogLines` I will get validation error if I keep it as `EmbeddedDocumentField`

`mongoengine.errors.ValidationError: ValidationError (Job:5d538097c2a21e95b5f8dc7c) (Invalid embedded document instance provided to an EmbeddedDocumentField: ['job_output'])`

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -528,13 +558,25 @@ def finish_job(self, job_id, ctx, error_message=None):
             job.errormsg = error_message
             self.get_mongo_util().update_job_status(job_id=job_id, status=Status.error.value)
         else:
+            if not job_output:
+                raise ValueError("Missing job output for finished job")
+
+            output = JobOutput()
+            output.version = job_output.get("version")
```
Tianhao-Gu: @bio-boris can you verify `output = self._watch(config)` does have 'version', 'id' and 'result' before calling finish_job?

https://github.com/kbase/JobRunner/blob/ee2/JobRunner/JobRunner.py#L292

lib/execution_engine2/models/models.py
bio-boris: Yea it's weird. JobInput is working right?

Tianhao-Gu: yup so strange. JobInput works.

PR #27, opened 2019-08-15
Fix logs

bio-boris:


lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -320,7 +325,7 @@ def __init__(self, config, ctx=None):
     def status():
         return {"servertime": f"{time()}"}
 
-    def cancel_job(self, job_id, ctx):
+    def cancel_job(self, job_id, ctx, terminated_code=None):
```
Tianhao-Gu: is `terminated_code` used anywhere?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -608,13 +618,16 @@ def get_job_status(self, job_id, ctx):
 
         return returnVal
 
-    def finish_job(self, job_id, ctx, error_message=None, job_output=None):
+    def finish_job(
+        self, job_id, ctx, error_message=None, job_output=None, finish_code=None
```
Tianhao-Gu: also I didn't see `finish_code` is used anywhere

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -624,33 +637,42 @@ def finish_job(self, job_id, ctx, error_message=None, job_output=None):
...
+            output.id = ObjectId(job_output.get("id"))
             output.result = job_output.get("result")
-            output.validate()
-            job.job_output = output
-
-            self.get_mongo_util().update_job_status(job_id=job_id, status=Status.finished.value)
+            try:
+                output.validate()
+                job.job_output = output
+                job.status = Status.finished.value
```
Tianhao-Gu: redundant with `update_job_status` in next line.

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -624,33 +637,42 @@ def finish_job(self, job_id, ctx, error_message=None, job_output=None):
...
+                self.get_mongo_util().update_job_status(
+                    job_id=job_id, status=Status.finished.value
+                )
+            except Exception as e:
+                logging.info(e)
+                job.output = None
+                job.errormsg = f"Something was wrong with the output we got  + {str(e)}"
+                self.get_mongo_util().update_job_status(
+                    job_id=job_id, status=Status.error.value
```
Tianhao-Gu: if `output` is not validated, shouldn't we still honor the job as `finished`? because there is no `error_message` given.

lib/execution_engine2/utils/SDKMethodRunner.py
bio-boris: Not yet, but in the next PR

lib/execution_engine2/utils/SDKMethodRunner.py
bio-boris: Not yet, but in the next PR

lib/execution_engine2/utils/SDKMethodRunner.py
bio-boris: If the output object wasn't fully corretly created, something went wrong.

lib/execution_engine2/utils/SDKMethodRunner.py
bio-boris: I don't understand

Tianhao-Gu: `update_job_status` was called right after `job.status = Status.finished.value`

PR #28, opened 2019-08-19
add check job and check jobs

Tianhao-Gu:


execution_engine2.spec
```diff
@@ -186,64 +186,24 @@ module execution_engine2 {
...
-        job_states - states of jobs,
-        job_params - parameters of jobs,
-        check_error - this map includes info about errors happening during job checking.
+        job_states - states of jobs
     */
     typedef structure {
-        mapping<job_id, JobState> job_states;
-        mapping<job_id, RunJobParams> job_params;
-        mapping<job_id, JsonRpcError> check_error;
+        mapping<job_id, UnspecifiedObject> job_states;
```
bio-boris: I think this is fine for now, but I wonder if we should actually define what a job state should look like..

execution_engine2.spec
```diff
@@ -186,64 +186,24 @@ module execution_engine2 {
...
     } CheckJobsResults;
 
-    funcdef check_jobs(CheckJobsParams params) returns (CheckJobsResults)
-        authentication required;
+    funcdef check_jobs(list<job_id> job_ids) returns (CheckJobsResults) authentication required;
+    /*
+      Check job for all jobs in a given workspace
+    */
+    funcdef list_job_statuses(string workspace_id) returns (CheckJobsResults) authentication required;
```
bio-boris: maybe list_job_statuses should be, get_job_statuses_by_workspace() ?

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -706,3 +706,52 @@ def start_job(self, job_id, ctx, skip_estimation=True):
...
+        check_jobs: check and return job status for a given of list job_ids
+
+        """
+
+        logging.info("Start fetching status for jobs: {}".format(job_ids))
+
+        job_states = dict()
+
+        for job_id in job_ids:
+            job_states[job_id] = self.check_job(job_id, ctx)
```
bio-boris: I think we should make make check_jobs a private function and then do a manual mongoengine query  select job where job in [id1,id2,id3] 
I don't think we need to check permission for each job, because if you have permission to the workspace, you should have permission to all the jobs.

lib/execution_engine2/utils/SDKMethodRunner.py
```diff
@@ -706,3 +706,52 @@ def start_job(self, job_id, ctx, skip_estimation=True):
...
+        """
+        check_job: check and return job status for a given job_id
+
+        Parameters:
+        job_id: id of job
+        """
+        logging.info("Start fetching status for job: {}".format(job_id))
+
+        if not job_id:
+            raise ValueError("Please provide valid job_id")
```
bio-boris: I think we should probably filter the output of this at some point.. I don't think we want to return the entire 'output' object.. Or do we ? It might be super slow.

execution_engine2.spec
Tianhao-Gu: 👍

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍 added project field. I don't know if we want to by default filter out output. maybe we do. Think for now, it's fine returning whole Job.

lib/execution_engine2/utils/SDKMethodRunner.py
Tianhao-Gu: 👍 I added a check_permission params to by pass this check when calling from `check_workspace_jobs`

execution_engine2.spec
Tianhao-Gu: I agree. I want to leave the flexibility here for now. eventually we do want to let user know what exactly it returns once we finalize models.

lib/execution_engine2/utils/SDKMethodRunner.py
bio-boris: Maybe we should combine check_job into check_jobs into one function, and expect an array to always be passed in
Also instead of using get_mongo_util.get_job, we should write a new one called get_jobs() that expects an array, and does a WHERE * in [job_id1,jobid2] instead of WHERE * = job_id
Also instead of deleting the keys, you can filter the query https://stackoverflow.com/a/33271295

Tianhao-Gu: 👍 sure.

PR #35, opened 2019-09-05
Update documentation for testing in docker

bio-boris:


briehl:
I think tests are failing here because of the change to the docker-compose.yml. You might want to either change the image for the `ee2` block, or put in `docker build . -t kbase/execution_engine2:test` in the `before_install` block

briehl:
Or, like you have in the .travis.yml notes, just make that image and push it to Dockerhub. I can work on that.

briehl:
Putting the test image up on dockerhub fixed the build. Codacy doesn't want to restart, but I think that's ok.

PR #36, opened 2019-09-11
tweak check job return

Tianhao-Gu:


Tianhao-Gu:

PR #40, opened 2019-09-17
Added notes, description to various fields

briehl:
This is just an update to the notes and formatting of the execution_engine2.spec file (for now).

Really, I just wanted to document somewhere other than the code what each field is and what's available. I don't have the list of allowed projection strings yet, but I'm guessing they map onto the class names from `lib/execution_engine2/models/models.py`?

I also want to just throw out a recommendation here to change the returned name of "_id" (which is a mongoism/database-ism) to either "id" or "job_id". I'd prefer "job_id", since that's used throughout the rest of the ee2 API.

### DIFF
--- a/execution_engine2.spec
+++ b/execution_engine2.spec
@@ -56,7 +56,7 @@ module execution_engine2 {
     */
     funcdef list_config() returns (mapping<string, string>) authentication optional;
 
-    /* Returns the current running version of the NarrativeJobService as a semantic version string. */
+    /* Returns the current running version of the execution_engine2 servicve as a semantic version string. */
     funcdef ver() returns (string);
 
     /* Simply check the status of this service to see queue details */
--- a/execution_engine2.spec
+++ b/execution_engine2.spec
@@ -2,6 +2,7 @@ module execution_engine2 {
 
     /* @range [0,1] */
     typedef int boolean;
+
     /*
         A time in the format YYYY-MM-DDThh:mm:ssZ, where Z is either the
         character Z (representing the UTC timezone) or the difference
@@ -11,8 +12,20 @@ module execution_engine2 {
             2013-04-03T08:56:32Z (UTC time)
     */
     typedef string timestamp;
+
     /* A job id. */
     typedef string job_id;
+
+    /*
+        A structure representing the Execution Engine status
+        reboot_mode - if 1, then in the process of rebooting
+        stopping_mode - if 1, then in the process of stopping
+        running_tasks_total - number of total running jobs
+        running_tasks_per_user - mapping from user id to number of running jobs for that user
+        tasks_in_queue - number of jobs in the queue that are not running
+        config - the current service configuration (see also list_config())
+        git_commit - the Git commit string denoting the deployed version of EE2
+    */
     typedef structure {
         boolean reboot_mode;
         boolean stopping_mode;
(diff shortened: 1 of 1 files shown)


execution_engine2.spec
```diff
@@ -190,43 +216,120 @@ module execution_engine2 {
...
         list<string> projection;
     } CheckWorkspaceJobsParams;
     funcdef check_workspace_jobs(CheckWorkspaceJobsParams params) returns (CheckJobsResults) authentication required;
+
     typedef structure {
         job_id job_id;
     } CancelJobParams;
+
+    /*
+        Cancels a job. This results in the status becoming "terminated" with termination_code 0.
```
bio-boris: This is true right now. However, we should add a termination_code parameter so we can terminate it with

```
    terminated_by_user = 0
    terminated_by_admin = 1
    terminated_by_automation = 2
```

Should we update the documentation to say this now, or update it once that functionality is actually put in?

execution_engine2.spec
```diff
@@ -190,43 +216,120 @@ module execution_engine2 {
...
+        timestamp created;
+        timestamp updated;
+
+        timestamp estimating
+        timestamp running;
+        timestamp finished;
+
+        int error_code;
+        string errormsg;
+
```
bio-boris: I think we also return an "error" object as well.

execution_engine2.spec
```diff
@@ -22,20 +35,26 @@ module execution_engine2 {
         mapping<string, string> config;
         string git_commit;
     } Status;
+
+    /* Returns the service configuration, including URL endpoints and timeouts */
```
bio-boris: Should we add
```
        public_keys = ['external-url', 'kbase-endpoint', 'workspace-url', 'catalog-url',
                       'shock-url', 'handle-url', 'auth-service-url', 'auth-service-url-v2',
                       'auth-service-url-allow-insecure',
                       'scratch', 'executable', 'docker_timeout',
                       'intialdir', 'transfer_input_files']
```

execution_engine2.spec
```diff
@@ -22,20 +35,26 @@ module execution_engine2 {
...
     funcdef list_config() returns (mapping<string, string>) authentication optional;
-    /* Returns the current running version of the NarrativeJobService. */
+
+    /* Returns the current running version of the NarrativeJobService as a semantic version string. */
```
bio-boris: Oh yea its ee2 now not njs

execution_engine2.spec
```diff
@@ -97,47 +118,45 @@ module execution_engine2 {
...
 
     funcdef update_job_status(UpdateJobStatusParams params) returns (job_id job_id)
         authentication required;
+
+    /*
+        line - string - a string to set for the log line.
+        is_error - int - if 1, then this line should be treated as an error, default 0
+        ts - string - a timestamp for the log line (optional)
+
+        @optional ts
```
bio-boris: Does the @optional keyword do anything?

bio-boris:
Looks great! We needed this! Provided some light feedback.

execution_engine2.spec
briehl: It's mostly a keyword for workspace objects, but it's generally in the KIDL language to mark a field as optional. Maybe one day in the far future when validation is automagically done, it'll be more necessary.

execution_engine2.spec
briehl: Sure. And can we fix the typo in `intialdir`? Maybe not in this PR, though, seems a touch out of scope.

execution_engine2.spec
briehl: Let's try to keep the doc up to date with the code.

execution_engine2.spec
briehl: Done, and respelled as "initial_dir" - that's what's used in the actual config file. I'll make a separate PR with a code update.

execution_engine2.spec
briehl: Fixed

execution_engine2.spec
briehl: Fixed. Just used the JsonRpcError object that's there. It has the same structure as the model.

bio-boris:
@Tianhao-Gu Are you going to review or can I merge?

Tianhao-Gu:
@bio-boris @briehl  looks good to me.

PR #42, opened 2019-09-18
Fix travis tests

briehl:
Mostly to verify that pulling the test image works as intended, and still tests against code changes to the repo.

I'll throw in some tests that should fail. So if Travis tests here fail, then all is well.


briehl:
Verified that forcing a test to fail crashes.

Also verified that adding failing code (in models.py) that would crash, causes tests to fail.

Reverting back to everything being happy = happiness all around.

briehl:
Yup, all is well, nothing to see here. Closing this PR.

Issue #47, opened 2019-09-18
Abstract away DB-specific data


bio-boris:
Job ids shouldn't be just called "_id", and users shouldn't be required to use Mongo clients to extract creation dates. Users shouldn't have to know anything about how the service is implemented in order to use data it provides.

PR #51, opened 2019-09-18
convert _id to job_id in returned job state

briehl:
Works toward #47 , specifically converts _id in the returned job state to "job_id", and adds reasonably exhaustive validation tests to the non-projected job_state structure.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -4,6 +4,7 @@ from dotenv import load_dotenv
 import pathlib
 from shutil import copyfile
 from execution_engine2.models.models import Job, JobInput, Meta
+from dateutil import parser as dateparser
 
 
 def get_example_job():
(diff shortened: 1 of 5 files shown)


test/test_utils.py
```diff
@@ -52,3 +53,118 @@ def bootstrap():
     if not os.path.exists(test_env):
         copyfile(f"{pwd}/test/env/{test_env}", f"{test_env}")
     load_dotenv("test.env", verbose=True)
+
+
+def validate_job_state(state):
```
bio-boris: I like this function. I wonder if we should use it outside of testing as well, when we return the job state.

briehl: I think it's fine not to. If we just use it to capture all the cases where Mongo data can get mangled during testing, then using it in production would just be redundant (and would slow down things that need to be fast).

briehl:
Codacy doesn't seem to want to rescan my changes, even though I fixed its issue (had a bare `except:` clause without a specific error to catch).

Issue #43, opened 2019-09-18
Add additional fields to status endpoint


bio-boris:
```
{"gitcommithash":"a2ed4cb88120aeb10a295919cb0fba85e13d462d","servertime":1568823552860,"service":"Notification Feeds Service","version":"1.0.2"}
```

@briehl

briehl:
Yup. Just need to make those pieces of info available. Shouldn't be hard.

There's others in the .spec file, but they might be overkill? Or maybe should be separate endpoints, like getting the queued job counts for all users. Though it might make sense to have the total queue state in this call (or maybe a separate one).

PR #54, opened 2019-09-19
add gitcommit to status, explicit tests on server methods

briehl:
Gonna add more status to this one and tweak the spec file around it.


briehl:
Eventually it would be nice to add other fields, so I left them as #TODOs in the .spec file, so I can move on to other things.

briehl:
Solves #43

bio-boris:
Looks good

PR #55, opened 2019-09-20
Some structural refactoring

briehl:
- [x] migrate database-specific components to the db directory
- [x] Make SDKMethodRunner it's own top level module
- Separate concerns from SDKMethodRunner - it should just run SDKMethods
  - [x] Create a separate auth roles module / auth client for fetching auth roles and caching them
  - Create a workspace_util module to handle workspace auth strategy
- [x] Rename utils.py -> slack_utils.py


briehl:
This is getting bigger than it should, so I'll request a review here.
Here's the main changes.

1. utils/util.py -> utils/slack_util.py - that module only has slack functions in it, so it should just be renamed. I'm not sure those are used anywhere, but they're a neat idea.

2. utils/MongoUtil.py and the various other database modules -> migrated to the db directory. Just to help keep things isolated and separated.

3. Move SDKMethodRunner.py from utils to root. It's not really a utility, it's the main runner module.

4. Extract admin role lookup code from SDKMethodRunner to an AuthUtil module. This makes use of the authclient's TokenCache to cache those roles. It also handles various lookup error cases that weren't handled before.

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -24,10 +24,13 @@
     ErrorCode,
 )
 from execution_engine2.utils.Condor import Condor
-from execution_engine2.utils.MongoUtil import MongoUtil
+from execution_engine2.db.MongoUtil import MongoUtil
 from installed_clients.CatalogClient import Catalog
 from installed_clients.WorkspaceClient import Workspace
 from installed_clients.authclient import KBaseAuth
+from .utils.auth import AuthUtil
```
bio-boris: Can we do full paths here for easier imports for various different testing libraries?

lib/execution_engine2/utils/auth.py
```diff
@@ -0,0 +1,94 @@
...
+from json import JSONDecodeError
+from ..exceptions import AuthError
+
+IS_ADMIN = "EE2_ADMIN"
+NOT_ADMIN = "NOT_ADMIN"
+
+_admin_cache = TokenCache()
+
+
+class AuthUtil(object):
```
bio-boris: We probably don't need to be python2 compatible, so we don't have to import from object

bio-boris:
Tests pass, looks good to me!

lib/execution_engine2/SDKMethodRunner.py
briehl: Sure, fixed.

lib/execution_engine2/utils/auth.py
briehl: Yup, thanks for catching that. Just a habit.

PR #56, opened 2019-09-23
Add check_jobs_date_range

bio-boris:


bio-boris:
@eapearson Do you mind taking a look at the spec file?

bio-boris:
# comments on ee2 spec:
​
just comments on methods I'll be using, which looks to be
​
- `check_job`
- `check_jobs`
- `check_jobs_date_range_for_user`
- `check_jobs_date_range_for_all`
- `get_job_logs`
- `cancel_job`

# Queue status
* Q: I don't see any methods for queue status. This is on the "verbal queue" for job browser features - if a user is watching a queued job they may want to see the state of that queue, and hopefully the position of their job within that queue.

* A: This currently exists in the condor-stats service but not in ee2. There are plans in the works to create an ADMIN DASHBOARD using the HTCOndor flask API. There can be endpoints which call htcondor directly with the python bindings, if we want to filter and create functions just for users and their dashboard.

# Documenting Error Codes
* Q: Perhaps I missed it, but it would be good if methods documented what common usage errors might be returned and their error codes. E.g. a non-admin user calling an admin method, referring to a non-existent job, etc.
* A: Created  https://github.com/kbase/execution_engine2/issues/59



## `check_job` & `check_jobs`
Q: To reduce the api footprint, these could be collapsed into just `check_jobs`. When designing the job browser bff service, I considered the usage of both singular and plural endpoints, and came down on the side of just supporting the plural endpoint. To me, the difference between the two methods is too slight.

A: I think the only thing that uses this is the narrative. @Tianhao-Gu  or @briehl  might know offhand which of these are used in the narrative and how important it is to have different methods.  The jobRunner doesn't use this AFAIK

(Unresolved)

# `check_jobs_date_range_for_user / all`
> ​
> This is the meat of the job browser functionality.
> ​
> I'm assuming this similar pair of methods is designed around a two-layered approach to supporting the job browser, in which the job browser uses the job browser service, which in turn uses these two methods.
> ​
> Humour me for a moment as I state the challenge here. For large result sets, that is for query conditions which result in a large candidate set of jobs, perhaps thousands, it is preferrable to utilize server side sort, search, and paging. And once you have made this design choice, you are all in! You also have think from the ultimate tool's, the ui's, perspective. Sorting and searching applies primarily to the exposed values. Thus, sorting and searching by narrative implies the narrative title, for apps implies the app title (although module name is handy too), etc. This means that these data sources need to be joined to the job data before filtering and sorting.
> ​
> In the current job browser, searching and sorting are primarily conducted on the front end. The job browser bff service exists in order to move this functionality from the front end to the back end. This should improve job browser performance and exapand it's functionality, especially for large sets of jobs. 
> ​
> So in this case, the `check_jobs...` methods are not called by the ui by by the job browser bff service. The bff service will need to fetch an entire batch of jobs, all jobs in the search space. Most of the check_jobs... methods return a limited number of jobs (check_job, check_jobs, check_workspace_jobs), but the ..date_range.. methods can return up to thousands of jobs with large time ranges. The implication is that the job browser bff service will fetch a possibly very large set of jobs, then perform the "join", then perform any filtering, then perform sorting, then extract the required subset specified by the offset & limit.
> ​
> To support scalable fetching of large data sets the ee2 methods need to support batching. For instance, if a call would result in 10,000 jobs, it may be more effective to make multiple parallel requests of some batch size, say 1,000. This can be accomplished with `offset` and `limit`.

> Given that some filtering and sorting will necessarily be conducted in the job browser bff service, I'm not sure there is much utility in doing filtering or sorting on the ee2 side. (This is in relation to the job browser needs - it isn't to say that other use cases couldn't use these features.) I mean, some sorting and filtering could be done on the ee2 side, which would reduce the payload from the ee2 api, and then additionally on the job browser bff side, but this may not be effective enough.

For example, I imagine that most job browse usage will involve first definiing a date range through one of the preselected ranges (day, 48 hr, week, month), and then perhaps some filtering or sorting applied. In other words, the job browser bff will already have fetched all of the data for a given time range. Once fetched, the job browser bff only needs to maintain that set of jobs by updating active jobs, a relatively lightweight operation. Sorting and filering will be in-memory operations and very fast.
​
Of course this design relies on some relatively low number of jobs.
​
​
### overall
​
- for batching purposes a consistent order is important; there doesn't need to be a sorting parameter but a sort order should be enforced by default and documented. E.g. ordered by creation time, by row id, etc.

# Queued
* Q: i don't see a `queued` (timestamp) field 
* A: (this is present in the job object, right now no apps are marked as queued , they go from created to running, as we still need the estimator function to determine what to do with the job)

# TODO 
* should have `offset` field, the paging partner to `limit`
* Resolve this: the projection documentation mentions fields which are not present in the JobState structure.
(Which ones are these?) (They are returning a "CheckJob" return, not a "Job" object)
* Resolve Question about checkjob and checkjob_s
* The return type should be a *list* not a *mapping* since it needs to be deterministically ordered to support paging / batching. 
* the return structure should include the `total_count`, the total # of jobs after time range (and filtering, if used) is applied, regardless of the specified limit. this supports a client making an initial request for a batch, and then able to schedule subsequent batches to get the rest.
* 
# Finished Status
* Q: I think there should be another term used for the `finished` status; this has always bugged me about the current job status - the term "finished" is used in the more general sense for the "finished" timestamp, but is used more specifically for "finished successfully" as a job status value. This has always been confusing. I think this would be cleared up by using one term to indicate the stopped condition of the job, and another for the "successfully completed" condition. I've been using `finished` in the more general sense (thus for the timestamp), and `completed` for the job status indicating "finished successfully". What do you think?- 
* A: We have new status codes at https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L208-L223 which address this


# Timestamp
 i'm not sure the `updated` timestamp is necessary - it seems that the job status informs which timestamps are relevant.
*The updated timestamp is simple the last time the record was touched through the API, which is good for very basic auditing purposes which reminds me, maybe we want to save the user associated with the request to cancel the job or do something to do the job somewhere in some audit trail  (see https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L198-L205)

# ERror Message
- `errormsg`: i'd suggest not suggesting a stack trace, but rather providing guidance that it id a short succinct, unformatted, single-line error message meant for humans. This allows the error message to be displayed to the end user more uniformly.-

- Then there is the error message, and error code,  which are more human readable, and  the "Error" object, which is the full strack trace and or error from the Job Runner. See the `models.py`

    errormsg = StringField()
    msg = StringField()
    error = DynamicField()

    terminated_code = IntField(validation=valid_termination_code)
    error_code = IntField(validation=valid_errorcode)


- it could be useful to have a termination message? E.g. admin or process termination should communicate to the user why the job was terminated, no? If finish/error/terminate were combined into one status, with a reason code, there could also be one finish message for either error or termination.
# TODO not sure how to answer this question
 



- i don't see the `client_group` (or whatever the terminology is in ee2) in the job state structure
https://github.com/kbase/execution_engine2/projects/1#card-26936084


## `get_job_logs`
TODO: generally this method should support paging
- the job browser bff service will simply pass through
​
### params
​
- should have paging parameters:
	- `offset`
	- `limit`
	
	and omit `skip_lines`. `skip_lines` would be simulated by just offset with no limit.
- should support search aka filter - a use case and the subject of at least one ticket is to find log entries within a very large log.
- filter by log "level" (is\_error)
- what do you think about log _levels_ vs just an `is_error` flag? I imagine this is either implemented or on the roadmap for job logs, it is pretty standard to log with a level and then allow filtering by log level. The front end and job browser service are implementing this, even if it is translated from the `is_error` flag for the interaction with the ee2 service.

 #A: This is complicated. We would have to have the developer specify this and or figure out the level based on REGEX??


- on a related note, do job logs have a `timestamp` that could be included? That is another common log feature that has been  mysteriously missing from the job logs. Perhaps not present in the stored logs?
* Yes, https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L78 



#TODO just to be consistent in terminology, instead of `last_line_number` how about `total_count`.

## `cancel_job`

* Q: is an error returned if the job is not in a cancellable state? This would seem useful. E.g. if a job finished before a user's cancellation request was serviced, it would be better to warn the user that the cancellation couldn't be carried out because they will see that the job is finished in the job browser, and they will see the job products in their narrative. 
* A Yes, see https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/MongoUtil.py#L200


This was pretty long. Some of this post can be made into seperate issues not related to this PR

eapearson:
> * Q: i don't see a `queued` (timestamp) field
> * A: (this is present in the job object, right now no apps are marked as queued , they go from created to running, as we still need the estimator function to determine what to do with the job)

I'm dense - I'm a little confused as to why there is no timestamp for when a job enters a queue.

eapearson:
> * what do you think about log _levels_ vs just an `is_error` flag? I imagine this is either implemented or on the roadmap for job logs, it is pretty standard to log with a level and then allow filtering by log level. The front end and job browser service are implementing this, even if it is translated from the `is_error` flag for the interaction with the ee2 service.
> 
> #A: This is complicated. We would have to have the developer specify this and or figure out the level based on REGEX??
> 

I don't know how much refactoring is going on in ee2, so I just wanted to raise this issue. To me, having just "is_error" as a flag for a log entry begs the issue of whether this should rather be a logging "level" so the user can select which types of log messages they are interested in viewing. It could help users cut through a lot of logging noise, allow the distinction between warnings and errors (the subject of at least one user issue I've seen reported) in a structured manner.

I would imagine that it would be implemented as a log level assigned when an app issues a log entry.

eapearson:
> # comments on ee2 spec:
> ​
> just comments on methods I'll be using, which looks to be
> ​
> 
> * `check_job`
> * `check_jobs`
> * `check_jobs_date_range_for_user`
> * `check_jobs_date_range_for_all`
> * `get_job_logs`
> * `cancel_job`
> 
> # Queue status
> * Q: I don't see any methods for queue status. This is on the "verbal queue" for job browser features - if a user is watching a queued job they may want to see the state of that queue, and hopefully the position of their job within that queue.
> * A: This currently exists in the condor-stats service but not in ee2. There are plans in the works to create an ADMIN DASHBOARD using the HTCOndor flask API. There can be endpoints which call htcondor directly with the python bindings, if we want to filter and create functions just for users and their dashboard.
> 
> # Documenting Error Codes
> * Q: Perhaps I missed it, but it would be good if methods documented what common usage errors might be returned and their error codes. E.g. a non-admin user calling an admin method, referring to a non-existent job, etc.
> * A: Created  #59
> 
> ## `check_job` & `check_jobs`
> Q: To reduce the api footprint, these could be collapsed into just `check_jobs`. When designing the job browser bff service, I considered the usage of both singular and plural endpoints, and came down on the side of just supporting the plural endpoint. To me, the difference between the two methods is too slight.
> 
> A: I think the only thing that uses this is the narrative. @Tianhao-Gu or @briehl might know offhand which of these are used in the narrative and how important it is to have different methods. The jobRunner doesn't use this AFAIK
> 
> (Unresolved)
> 
> # `check_jobs_date_range_for_user / all`
> > ​
> > This is the meat of the job browser functionality.
> > ​
> > I'm assuming this similar pair of methods is designed around a two-layered approach to supporting the job browser, in which the job browser uses the job browser service, which in turn uses these two methods.
> > ​
> > Humour me for a moment as I state the challenge here. For large result sets, that is for query conditions which result in a large candidate set of jobs, perhaps thousands, it is preferrable to utilize server side sort, search, and paging. And once you have made this design choice, you are all in! You also have think from the ultimate tool's, the ui's, perspective. Sorting and searching applies primarily to the exposed values. Thus, sorting and searching by narrative implies the narrative title, for apps implies the app title (although module name is handy too), etc. This means that these data sources need to be joined to the job data before filtering and sorting.
> > ​
> > In the current job browser, searching and sorting are primarily conducted on the front end. The job browser bff service exists in order to move this functionality from the front end to the back end. This should improve job browser performance and exapand it's functionality, especially for large sets of jobs.
> > ​
> > So in this case, the `check_jobs...` methods are not called by the ui by by the job browser bff service. The bff service will need to fetch an entire batch of jobs, all jobs in the search space. Most of the check_jobs... methods return a limited number of jobs (check_job, check_jobs, check_workspace_jobs), but the ..date_range.. methods can return up to thousands of jobs with large time ranges. The implication is that the job browser bff service will fetch a possibly very large set of jobs, then perform the "join", then perform any filtering, then perform sorting, then extract the required subset specified by the offset & limit.
> > ​
> > To support scalable fetching of large data sets the ee2 methods need to support batching. For instance, if a call would result in 10,000 jobs, it may be more effective to make multiple parallel requests of some batch size, say 1,000. This can be accomplished with `offset` and `limit`.
> 
> > Given that some filtering and sorting will necessarily be conducted in the job browser bff service, I'm not sure there is much utility in doing filtering or sorting on the ee2 side. (This is in relation to the job browser needs - it isn't to say that other use cases couldn't use these features.) I mean, some sorting and filtering could be done on the ee2 side, which would reduce the payload from the ee2 api, and then additionally on the job browser bff side, but this may not be effective enough.
> 
> For example, I imagine that most job browse usage will involve first definiing a date range through one of the preselected ranges (day, 48 hr, week, month), and then perhaps some filtering or sorting applied. In other words, the job browser bff will already have fetched all of the data for a given time range. Once fetched, the job browser bff only needs to maintain that set of jobs by updating active jobs, a relatively lightweight operation. Sorting and filering will be in-memory operations and very fast.
> ​
> Of course this design relies on some relatively low number of jobs.
> ​
> ​
> 
> ### overall
> ​
> 
> * for batching purposes a consistent order is important; there doesn't need to be a sorting parameter but a sort order should be enforced by default and documented. E.g. ordered by creation time, by row id, etc.
> 
> # Queued
> * Q: i don't see a `queued` (timestamp) field
> * A: (this is present in the job object, right now no apps are marked as queued , they go from created to running, as we still need the estimator function to determine what to do with the job)
> 
> # TODO
> * should have `offset` field, the paging partner to `limit`
> * Resolve this: the projection documentation mentions fields which are not present in the JobState structure.
>   (Which ones are these?) (They are returning a "CheckJob" return, not a "Job" object)
> * Resolve Question about checkjob and checkjob_s
> * The return type should be a _list_ not a _mapping_ since it needs to be deterministically ordered to support paging / batching.
> * the return structure should include the `total_count`, the total # of jobs after time range (and filtering, if used) is applied, regardless of the specified limit. this supports a client making an initial request for a batch, and then able to schedule subsequent batches to get the rest.
> 
> # Finished Status
> * Q: I think there should be another term used for the `finished` status; this has always bugged me about the current job status - the term "finished" is used in the more general sense for the "finished" timestamp, but is used more specifically for "finished successfully" as a job status value. This has always been confusing. I think this would be cleared up by using one term to indicate the stopped condition of the job, and another for the "successfully completed" condition. I've been using `finished` in the more general sense (thus for the timestamp), and `completed` for the job status indicating "finished successfully". What do you think?-
> * A: We have new status codes at https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L208-L223 which address this
> 
> # Timestamp
> i'm not sure the `updated` timestamp is necessary - it seems that the job status informs which timestamps are relevant.
> *The updated timestamp is simple the last time the record was touched through the API, which is good for very basic auditing purposes which reminds me, maybe we want to save the user associated with the request to cancel the job or do something to do the job somewhere in some audit trail (see https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L198-L205)
> 
> # ERror Message
> * `errormsg`: i'd suggest not suggesting a stack trace, but rather providing guidance that it id a short succinct, unformatted, single-line error message meant for humans. This allows the error message to be displayed to the end user more uniformly.-
> * Then there is the error message, and error code,  which are more human readable, and  the "Error" object, which is the full strack trace and or error from the Job Runner. See the `models.py`
>   errormsg = StringField()
>   msg = StringField()
>   error = DynamicField()
>   terminated_code = IntField(validation=valid_termination_code)
>   error_code = IntField(validation=valid_errorcode)
> * Q: it could be useful to have a termination message? E.g. admin or process termination should communicate to the user why the job was terminated, no? If finish/error/terminate were combined into one status, with a reason code, there could also be one finish message for either error or termination.
> 
> * A: TODO not sure how to answer this question
> 
> * i don't see the `client_group` (or whatever the terminology is in ee2) in the job state structure
>   https://github.com/kbase/execution_engine2/projects/1#card-26936084
> 
> ## `get_job_logs`
> TODO: generally this method should support paging
> 
> * the job browser bff service will simply pass through
>   ​
> 
> ### params
> ​
> 
> * should have paging parameters:
>   
>   * `offset`
>   * `limit`
>   
>   and omit `skip_lines`. `skip_lines` would be simulated by just offset with no limit.
> * should support search aka filter - a use case and the subject of at least one ticket is to find log entries within a very large log.
> * filter by log "level" (is_error)
> * what do you think about log _levels_ vs just an `is_error` flag? I imagine this is either implemented or on the roadmap for job logs, it is pretty standard to log with a level and then allow filtering by log level. The front end and job browser service are implementing this, even if it is translated from the `is_error` flag for the interaction with the ee2 service.
> 
> #A: This is complicated. We would have to have the developer specify this and or figure out the level based on REGEX??
> 
> * on a related note, do job logs have a `timestamp` that could be included? That is another common log feature that has been  mysteriously missing from the job logs. Perhaps not present in the stored logs?
> 
> * Yes, https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L78
> 
> #TODO just to be consistent in terminology, instead of `last_line_number` how about `total_count`.
> 
> ## `cancel_job`
> * Q: is an error returned if the job is not in a cancellable state? This would seem useful. E.g. if a job finished before a user's cancellation request was serviced, it would be better to warn the user that the cancellation couldn't be carried out because they will see that the job is finished in the job browser, and they will see the job products in their narrative.
> * A Yes, see https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/MongoUtil.py#L200
> 
> This was pretty long. Some of this post can be made into seperate issues not related to this PR



> > * what do you think about log _levels_ vs just an `is_error` flag? I imagine this is either implemented or on the roadmap for job logs, it is pretty standard to log with a level and then allow filtering by log level. The front end and job browser service are implementing this, even if it is translated from the `is_error` flag for the interaction with the ee2 service.
> > 
> > #A: This is complicated. We would have to have the developer specify this and or figure out the level based on REGEX??
> 
> I don't know how much refactoring is going on in ee2, so I just wanted to raise this issue. To me, having just "is_error" as a flag for a log entry begs the issue of whether this should rather be a logging "level" so the user can select which types of log messages they are interested in viewing. It could help users cut through a lot of logging noise, allow the distinction between warnings and errors (the subject of at least one user issue I've seen reported) in a structured manner.
> 
> I would imagine that it would be implemented as a log level assigned when an app issues a log entry.

But I realize now that logging functionality is really just stdout and stderr, so further refinement of log level is not possible without additional changes.

eapearson:
I'd suggest we just get this up and iterate on it rather than let it linger. We need this api up in order to start working against it as a kb_Metrics replacement.

PR #58, opened 2019-09-25
Workspace auth strategy

briehl:
The goal here is to abstract the auth strategy lookup out of SDKMethodRunner for a couple reasons.
1. It separates the concerns out of that module. The method runner shouldn't care about the workspace or handling a workspace client.
2. If we ever get a different auth strategy, we can just plug it in without much effort.
3. We needed a way to trap errors that come from the WS service.


briehl:
@bio-boris @Tianhao-Gu 
This started to get more complex as I started adding tests and such, so I'm stopping this PR here.

This adds a few auth functions and tests around them. The main functions are in `execution_engine2/authorization/authstrategy.py` and they can be used to check user permissions for one job or a list of them. Both checks should take about the same time.

The next step (already started, in the next PR) is adapting SDKMethodRunner to use these to look them up. That should be much less complex, but there's also a bunch of testing to go on there, too. The benefit there, is that it should only involve one job lookup when it's done, instead of two (one for perms test, and one for the function).

For tests, this uses the `requests_mock` framework https://requests-mock.readthedocs.io/en/latest/ - this catches any http call made through `requests` (so, all of our http clients) and returns whatever we want it to. I think it's pretty snazzy, but can be tricky to set up.

Abstracting the authstrategy stuff might be a little overkill, but if we ever move on to not just using the workspace, then there's already a pattern there. But if you all think that's overengineered, it shouldn't be hard to undo it.

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -311,7 +316,12 @@ def __init__(self, config, ctx=None):
 
         self.workspace_url = config.get("workspace-url")
         self.auth_url = config.get("auth-url")
-        self.auth_util = AuthUtil(self.auth_url, self.admin_roles)
+        self.admin_auth_util = AdminAuthUtil(self.auth_url, self.admin_roles)
+        self.user_id = user_id
+        self.token = token
+        self.is_admin = False
+        if self.token:
+            AdminAuthUtil(self.auth_url, self.admin_roles).is_admin(self.token)
```
bio-boris: Do we always need to check to see if someone is an admin?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -482,6 +492,10 @@ def administer(self, command, params, token):
                 f"You are not authorized. Please request a role from {self.admin_roles}"
             )
 
+    def check_job_permission(self, job: Job):
```
bio-boris: Is this unused?

test/ee2_workspaceauth_test.py
```diff
@@ -0,0 +1,115 @@
+import unittest
```
bio-boris: Have you ever tried cassettes? You can specify the requests in yaml. I tried to use it at https://github.com/kbase/JobRunner/pull/4/files at some point.

lib/execution_engine2/authorization/authstrategy.py
```diff
@@ -0,0 +1,139 @@
...
+    #    of job indices as values.
+    # 3. The rest (not the same user, and a different authstrat) are not allowed, so leave them
+    #    marked False in the perms list
+    # 4. Now, if there's any ids in that wsid dict, use WorkspaceAuth to look 'em up.
+    # 5. We can now unravel the mappings. ws_auth_perms -> wsid -> list of indices of jobs with that
+    #    wsid.
+    # should be just a single ws call to get_permissions_mass, and up to 2 loops over the list of
+    # jobs (O(2N) if there's a unique ws id for each job, and fewer comparisons)
+
+    perms = len(jobs) * [False]    # permissions start False
```
bio-boris: The alternative could be to use a `defaultdict(lambda: False)` instead of this list

lib/execution_engine2/authorization/authstrategy.py
```diff
@@ -0,0 +1,139 @@
...
+        ws_auth = WorkspaceAuth(token, user_id, config['workspace-url'])
+        if level == "read":
+            return ws_auth.can_read(job.wsid)
+        else:
+            return ws_auth.can_write(job.wsid)
+    else:
+        return user_id == job.user
+
+
+def _check_permissions_list(jobs: List[Job], user_id: str, token: str, config: Dict[str, str], level="read") -> bool:
```
bio-boris: I think the return type is not a bool, but a list of bool

lib/execution_engine2/authorization/authstrategy.py
bio-boris: or `defaultdict(bool)`?

lib/execution_engine2/SDKMethodRunner.py
briehl: Not necessarily. I figure that admins just get superuser powers (can view/modify any job, can run admin functions) that can just skirt around the other auth mechanisms. But we can just put that in the general auth checking stages, too, as necessary.

lib/execution_engine2/SDKMethodRunner.py
briehl: Oops, yeah that's meant for the next PR. I'll remove it.

test/ee2_workspaceauth_test.py
briehl: Were you able to get it to work? That all looks commented out.

lib/execution_engine2/authorization/authstrategy.py
briehl: That's the right thing to do. Dangit, I always forget about defaultdict...

lib/execution_engine2/authorization/authstrategy.py
briehl: Yes. Thanks for catching that.

lib/execution_engine2/authorization/authstrategy.py
briehl: Oh, I thought you meant below where it checks for dict initialization. That list needs to be a fixed size, matching the number of jobs.

lib/execution_engine2/SDKMethodRunner.py
briehl: Removed.

lib/execution_engine2/authorization/authstrategy.py
briehl: Updated to defaultdict

lib/execution_engine2/SDKMethodRunner.py
briehl: Ok, removed that for now. the `administer` function is the only place that checks admin rights now. I think other changes (converting to using authstrategy for rights checking) will be the next change set.

test/ee2_workspaceauth_test.py
bio-boris: Yes, but I just didn't have that much use for it in the job runner and using it got rejected

briehl: I think in this case, I prefer requests_mock. It's maybe a little more code to set up, but it's pretty easy to parameterize and set up handlers to craft responses based on test input. It's also useful to set up explicit error conditions that might be hard to catch with real server calls (like timeouts, bad or incomplete data, etc.)

PR #63, opened 2019-09-30
Transfer

bio-boris:


lib/execution_engine2/db/models/transfer/transfer_logs.py
```diff
@@ -0,0 +1,110 @@
...
+            job_log.original_line_count = log['original_line_count']
+            job_log.stored_line_count = log['stored_line_count']
+
+            lines = []
+            for line in log['lines']:
+                ll = LogLines()
+                ll.error = line['is_error']
+                ll.linepos  = line['line_pos']
+                ll.line = line['line']
+                ll.validate()
```
Tianhao-Gu: It will set LogLines tz all to the current time. Is it expected?

bio-boris: Yea we don't have a default time for these logs so I think that's fine. Or we can explicitly set a time, like "january 1, 1970". What do you think we should do?

Tianhao-Gu: 👍 cool. default to current time looks good to me.

PR #65, opened 2019-10-03
Admin console issues

briehl:
Admin access should only be checked when using the administer command, and shouldn't override any other failed permissions. That is, a user who happens to have an admin role shouldn't just be able to run cancel_job on any job - that seems like an easy way to accidentally cancel things.


briehl:
Note that this builds on PR #64 , so that should be dealt with and merged first.

Issue #69, opened 2019-10-07
Timestamp format inconsistent and not compliant with the spec


eapearson:
ee2 job state timestamp fields should be consistent with the definition in the spec file, but are not. Some include sub second time, most don't have the tz offset, none have the "T" separator between date and time, and probably the spec should reference the actual standard (ISO8601 by the look of it). 

The spec states:

```
 A time in the format YYYY-MM-DDThh:mm:ssZ, where Z is either the
        character Z (representing the UTC timezone) or the difference
        in time to UTC in the format +/-HHMM, eg:
            2012-12-17T23:24:06-0500 (EST time)
            2013-04-03T08:56:32+0000 (UTC time)
            2013-04-03T08:56:32Z (UTC time)
```

E.g.  this is snipped from the results of get_job in CI:

```
 "updated": "2019-09-09 17:52:27.380000",
 "running": "2019-09-09 17:47:38",
 "finished": "2019-09-09 17:52:27",
 "created": "2019-09-09 17:47:04+00:00"
```

Tianhao-Gu:
@eapearson 
format record job timestamp: https://github.com/kbase/execution_engine2/pull/75

'created': '2019-10-16T21:04:40.559732+00:00'

Tianhao-Gu:
per https://github.com/kbase/execution_engine2/pull/84. all of the time fields should be in epoch format now.

scanon:
This is fixed.

PR #75, opened 2019-10-16
format time

Tianhao-Gu:


Tianhao-Gu:
address issue https://github.com/kbase/execution_engine2/issues/69

Tianhao-Gu:
@bio-boris Do you mind take a look?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -746,14 +746,14 @@ def check_jobs(self, job_ids, check_permission=True, projection=None):
             mongo_rec = job.to_mongo().to_dict()
             del mongo_rec['_id']
             mongo_rec['job_id'] = str(job.id)
-            mongo_rec['created'] = str(job.id.generation_time)
-            mongo_rec['updated'] = str(job.updated)
+            mongo_rec['created'] = job.id.generation_time.utcnow().replace(tzinfo=timezone.utc).isoformat()
```
bio-boris: Can you just do   job.generation_time.isoformat() ? What is the the utcnow.replace for?

bio-boris:
Looks good. Let's get it up on CI.

eapearson:
Won't this change set all dates to the current time in UTC? That is what utcnow() does. 

Also, the stock isoformat() method will output datetime strings like
 '2019-10-23T13:01:07.344000+00:00'. 

This is not compliant with the format stated in the spec:
```
 A time in the format YYYY-MM-DDThh:mm:ssZ, where Z is either the
        character Z (representing the UTC timezone) or the difference
        in time to UTC in the format +/-HHMM, eg:
            2012-12-17T23:24:06-0500 (EST time)
            2013-04-03T08:56:32+0000 (UTC time)
            2013-04-03T08:56:32Z (UTC time)
```

It might save some time now and going forward if the datetime values were simply epoch time in ms, as most other apis do.

bio-boris:
Yes you are right @eapearson 

```
import datetime
from datetime import timedelta, timezone

now = datetime.datetime.utcnow()
last_week = now - timedelta(days=7)

a = last_week
b = last_week.replace(tzinfo=timezone.utc).isoformat()
c = last_week.utcnow().replace(tzinfo=timezone.utc).isoformat()
d = last_week.isoformat()

print(a)
print(b)
print(c)
print(d)

2019-10-16 20:47:37.661041
2019-10-16T20:47:37.661041+00:00
2019-10-23T20:47:37.661072+00:00
2019-10-16T20:47:37.661041
```

bio-boris:
Ah you are right, its +0000, not +00:00

I don't remember why anymore why we chose using a timestamp vs epoch, but we did discuss that at some point. But that is an option, and I'm OK with doing that.

PR #84, opened 2019-10-24
switching to epoch time

Tianhao-Gu:


Tianhao-Gu:
@bio-boris huh? I already changed models from `DateTimeField` to `FloatField`.

bio-boris:
@Tianhao-Gu Sorry I was looking at https://github.com/kbase/execution_engine2/blob/fa7620d9465be1ee88fadd718e72c9562df138ce/lib/execution_engine2/db/models/models.py#L80

I think that's the only one that is left.

bio-boris:
Other than that one last log line, looks great!

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -280,8 +280,14 @@ def add_job_logs(self, job_id, log_lines):
...
             if ts is not None:
-                if type(ts) == str:
-                    ts = dateutil.parser.parse(ts)
+                try:     # input ts as float/int/str epoch timestamp
+                    ts = float(ts)
+                except Exception:
+                    try:  # input ts as datetime string
+                        ts = dateutil.parser.parse(ts).timestamp()
+                    except Exception:
+                        ts = None  # TODO raise error?
```
bio-boris: You could set TS to now instead of setting it to None

bio-boris: Not sure if that would cause problems though with log ordering.... But I think the ordering is done on the position, not the timestamp..

Tianhao-Gu: 👍 I think `LogLines` will default ts to utcnow() if `None`. But it doesn't hurt.

PR #86, opened 2019-10-30
Use epoch time

Tianhao-Gu:
per @eapearson we should use epoch time in milliseconds.

### DIFF
--- a/lib/execution_engine2/SDKMethodRunner.py
+++ b/lib/execution_engine2/SDKMethodRunner.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 from datetime import datetime, timezone
+import time
 from enum import Enum
 
 import dateutil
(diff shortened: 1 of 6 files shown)


lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -280,13 +281,24 @@ def add_job_logs(self, job_id, log_lines):
             ts = input_line.get("ts")
             # TODO Maybe use strpos for efficiency?
             if ts is not None:
-                try:     # input ts as float/int/str epoch timestamp
-                    ts = float(ts)
-                except Exception:
-                    try:  # input ts as datetime string
-                        ts = dateutil.parser.parse(ts).timestamp()
+                if isinstance(ts, int):  # input ts as epoch in milliseconds
```
bio-boris: It's most likely going to be a string, so maybe put that as the first if?
Should we have test cases for each of these instances?

Tianhao-Gu: 👍 sounds good. Added different input time formats in the test.

test/test_utils.py
```diff
@@ -163,13 +163,13 @@ def validate_job_state(state):
...
 
-def is_timestamp(ts: float):
+def is_timestamp(ts: int):
     """
     Simple enough - if dateutil.parser likes the string, it's a time string and we return True.
     Otherwise, return False.
     """
     try:
-        datetime.fromtimestamp(ts)
+        datetime.fromtimestamp(ts / 1000.0)
```
bio-boris: probably don't need this

PR #91, opened 2019-11-05
Fix timestamp bugfix

bio-boris:


lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -862,7 +862,7 @@ def _job_state_from_jobs(jobs):
             mongo_rec = job.to_mongo().to_dict()
             mongo_rec["_id"] = str(job.id)
             mongo_rec["job_id"] = str(job.id)
-            mongo_rec["created"] = str(job.id.generation_time)
+            mongo_rec["created"] = str(job.id.generation_time.timestamp())
             mongo_rec["updated"] = str(job.updated)
```
Tianhao-Gu: Do we want to return float string or just float here? I would prefer to return float value. Unless there is a requirement from Eric.

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -857,19 +871,31 @@ def check_workspace_jobs(self, workspace_id, projection=None):
...
-                mongo_rec["estimating"] = str(job.estimating)
+                mongo_rec["estimating"] = int(job.estimating) * 1000
+            if job.queued:
+                mongo_rec["queued"] = int(job.queued) * 1000
             if job.running:
-                mongo_rec["running"] = str(job.running)
+                mongo_rec["running"] = int(job.running) * 1000
             if job.finished:
-                mongo_rec["finished"] = str(job.finished)
+                mongo_rec["finished"] = int(job.finished) * 1000
```
Tianhao-Gu: It will round timestamp to whole seconds. I don't know if it matters but we should do int(timestamp() * 1000).

bio-boris: Thanks I fixed this

PR #95, opened 2019-11-06
fix run_time

Tianhao-Gu:


Tianhao-Gu:
`running` time field is incorrect due to this. 

created: 1573070985331
finished: 1573070984261
running: 1573070953678000
updated: 1573070984261

PR #97, opened 2019-11-07
Cancel job with bindings

bio-boris:


lib/execution_engine2/db/MongoUtil.py
```diff
@@ -211,17 +211,17 @@ def cancel_job(self, job_id=None, terminated_code=None):
...
-        :param job: Cancel job with actual job document instance
         """
 
         with self.mongo_engine_connection():
             j = self.get_job(job_id, projection=None)
             self.check_if_already_finished(j.status)
             if terminated_code is None:
                 terminated_code = TerminatedCode.terminated_by_user.value
-
+            j.terminated = time.time()
```
Tianhao-Gu: model doesn't need to be updated to include this field?

bio-boris: That gets fixed https://github.com/kbase/execution_engine2/pull/101/files#r345449369m, ill fix it here too

PR #101, opened 2019-11-11
Ran on ci, and saw a bug with cancel

bio-boris:


lib/execution_engine2/db/MongoUtil.py
```diff
@@ -222,6 +221,7 @@ def cancel_job(self, job_id=None, terminated_code=None):
             if terminated_code is None:
                 terminated_code = TerminatedCode.terminated_by_user.value
 
+            j.finished = time.time()
```
bio-boris: j.finished = time.time()

Issue #104, opened 2019-11-12
For get_job_logs, the ts timestamp is in seconds with decimal portion not milliseconds


eapearson:
Here is an sample of a log entry returned by a call to execution_engine2.get_job_logs:

```
 {
                    "line": "Job runner recieved Callback URL http://10.42.214.9:34919/",
                    "linepos": 1,
                    "error": false,
                    "ts": 1573236722.503325
                },
```

bio-boris:
@eapearson Can we close this ticket

bio-boris:
@eapearson Closing this

PR #105, opened 2019-11-12
fix get_job_log ts

Tianhao-Gu:
fix issue reported by https://github.com/kbase/execution_engine2/issues/104

### DIFF
--- a/lib/execution_engine2/SDKMethodRunner.py
+++ b/lib/execution_engine2/SDKMethodRunner.py
@@ -199,7 +199,7 @@ class SDKMethodRunner:
                     "line": log_line.line,
                     "linepos": log_line.linepos,
                     "error": log_line.error,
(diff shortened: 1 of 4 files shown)


lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -199,7 +199,7 @@ def _get_job_log(self, job_id, skip_lines):
                     "line": log_line.line,
                     "linepos": log_line.linepos,
                     "error": log_line.error,
-                    "ts": log_line.ts,
+                    "ts": int(log_line.ts * 1000),
```
bio-boris: If it is in MS, why do we have to multiply it?

bio-boris: Oh sorry nevermind

Issue #80, opened 2019-10-23
check_jobs* family of methods should return list not map


eapearson:
Since the check_jobs_date_range_for_* will be paged (still needs offset to implement this), the results need to be a sorted list. I believe that issue is in a separate ticket. Currently the results are a map, with the job id as the key and the job state info as the value.

https://github.com/kbase/execution_engine2/blob/8142da9bea25aa5f3acf0faf5115dce2b24af5fc/execution_engine2.spec#L321

For consistency, check_jobs_date_range_for_user and check_jobs_date_range_for_all return a list, then check_jobs perhaps should as well. In the case of check_jobs, the order should be the order in which the job_ids were provided.

Tianhao-Gu:
@eapearson Due to narrative backward compatibility, I think we have to keep check_jobs the way it is now. Since narrative cache old job states. We can change check_jobs_date_range_for_ the same as check_jobs if you prefer.

bio-boris:
So this ticket is 4 parts, here is what's done

- [ ]  check_jobs_date_range_for_user return a list
- [ ]  check_jobs_date_range_for_all return a list
- [ ]  check_jobs return a list
- [ ]  check_jobs return a list in order provided

Tianhao-Gu:
@bio-boris @briehl I think check_jobs cannot return a list due to backward compatibility in narrative. Otherwise old jobs won't work in narrative due to cache?

Tianhao-Gu:
or we can make it return as list as an option.

bio-boris:
@eapearson 
Can you please confirm
So for jobs 100, 9000, 200, 500 
you would expect
[ {100}, {9000}, {200}, {500} ] 
And if a record was missing youd get
[ {100}, {empty}, {200}, {500} ]

Tianhao-Gu:
closing this issue. @eapearson feel free to let me know if you still see this as an issue.

Issue #49, opened 2019-09-18
Describe & Document job states


bio-boris:
Job states shouldn't just be UnspecifiedObjects, but should be well-described. This is a pretty core piece of the API, and consuming applications should know what they're getting.

scanon:
@bio-boris Is this done?

bio-boris:
Yes

PR #106, opened 2019-11-12
check_jobs return list by default

Tianhao-Gu:
make check_jobs return a sorted list by default
https://github.com/kbase/execution_engine2/issues/80


execution_engine2.spec
```diff
@@ -314,29 +314,37 @@
 
         /*
             job_states - states of jobs
+
+            could be mapping<job_id, JobState> or list<JobState>
         */
         typedef structure {
-            mapping<job_id, JobState> job_states;
+            UnspecifiedObject job_states;
```
bio-boris: Shouldn't list be a  list<JobStates> ?

bio-boris: Oh I see.. It can return both...

bio-boris: But then this breaks what @briehl said in https://github.com/kbase/execution_engine2/issues/49

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -835,9 +835,14 @@ def check_jobs(self, job_ids, check_permission=True, projection=None):
 
             job_states[str(job.id)] = mongo_rec
 
+        job_states = dict(sorted(job_states.items()))
+
+        if return_list:
+            job_states = list(job_states.values())
```
bio-boris: What would happen if the string "False" was passed in here?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -835,9 +835,14 @@ def check_jobs(self, job_ids, check_permission=True, projection=None):
 
             job_states[str(job.id)] = mongo_rec
 
+        job_states = dict(sorted(job_states.items()))
```
bio-boris: Does this return in the order that the jobs  were requested? For example

1000,500,300 would return the records as 1000,500,300 and not as 300,500,1000

lib/execution_engine2/SDKMethodRunner.py
Tianhao-Gu: 👍 hopefully, no one does that. I found your `parse_bool_from_string`. 💯

lib/execution_engine2/SDKMethodRunner.py
Tianhao-Gu: 💯 added sort_by to `get_jobs`

execution_engine2.spec
Tianhao-Gu: yea. I updated it as `list` since it's the default return type now.

Tianhao-Gu:
@bio-boris Could you take another look?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -838,9 +838,12 @@ def check_jobs(self, job_ids, check_permission=True, projection=None):
 
             job_states[str(job.id)] = mongo_rec
 
+        if return_list is not None and SDKMethodRunner.parse_bool_from_string(return_list):
+            job_states = list(job_states.values())
```
bio-boris: I think you need to iterate over the job ids, like 100,900,200,9999 and return them in that order like
list = [{job_state_for_id_100}, {job_state_for_id_900}, {'maybe this job state is missing'},{job_state_for_id_999}]

lib/execution_engine2/SDKMethodRunner.py
bio-boris: Ah did you use it? I swear I saw you use it in a diff commit and now its gone from this PR or maybe its just too late at night right now.

lib/execution_engine2/SDKMethodRunner.py
Tianhao-Gu: @bio-boris I added a `order_by` in `get_jobs`. So the default jobs returned are sorted by id.

lib/execution_engine2/SDKMethodRunner.py
Tianhao-Gu: @bio-boris hmmm no it's there. `if return_list is not None and SDKMethodRunner.parse_bool_from_string(return_list):`

Tianhao-Gu:
@bio-boris check_jobs should now return jobs in same order as the input.

PR #111, opened 2019-11-13
Cancel and sigterm logging and fix spec

bio-boris:


bio-boris:
Yea I think its some bad state transitions in here too. Jobs started not displaying correctly after testing this on CI.

lib/execution_engine2/utils/Condor.py
```diff
@@ -353,7 +353,7 @@ def cancel_job(self, job_id: str) -> bool:
         :param job_id:
         :return:
         """
-        return self.cancel_jobs([job_id])
+        return self.cancel_jobs(["job_id"])
```
Tianhao-Gu: I think this line shouldn't be changed.

bio-boris: @Tianhao-Gu Fixed it thanks. Condor doesn't like` ints`

PR #127, opened 2019-11-18
add some load test

Tianhao-Gu:


Tianhao-Gu:
@bio-boris I think it has most of what I want to test. It pretty much covers init jobs, get jobs and update jobs into different status.

test/ee2_load_test.py
```diff
@@ -0,0 +1,195 @@
...
+            ],
+            "source_ws_objects": ["a/b/c", "e/d"],
+            "parent_job_id": "9998",
+            "meta": {"tag": "dev", "token_id": "12345"},
+        }
+
+        return job_params
+
+    def test_init_job_stress(self):
+        with self.mongo_util.mongo_engine_connection():
```
bio-boris: Can you please add the description of what they do in a comment.

bio-boris:
This is a good test for race conditions and such as a unit test.

Can we also please get a unit test that called the actual CI endpoints using the clients for

* check_job
* check_job_canceled
* run_job
* get_job_logs

I think if you do all of these it won't be much more work to do most of the funcdefs on https://github.com/kbase/execution_engine2/blob/develop/execution_engine2.spec

Tianhao-Gu:
@bio-boris do you mind giving it another review?

PR #130, opened 2019-11-19
Added Kafka + code reformat

bio-boris:
* Provides a utility function to log messages
* TODO: Configuration Stuff


lib/execution_engine2/utils/KafkaUtilsLogger.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+"""Module to provide kafka handlers for internal logging facility."""
```
bio-boris: This file currently unused

bio-boris:
```
I think it should be an event logger class and you should instantiate it
```
Do we have an example of this?
Do we want to use the python logging module to send messages to this? If so, I have an example utilities file to do it in this PR...

```
You may want to abstract it a little more
```
Could you please elaborate? 
Like create a bunch of helper functions that take keywords and generate a message and send based on that?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -389,12 +397,25 @@ def cancel_job(self, job_id, terminated_code=None):
...
+        }
+        send_message_to_kafka(data=cancel_job_msg)
+
         logging.info(f"About to cancel job in CONDOR using {job.scheduler_id}")
         logging.debug(f"About to cancel job in CONDOR using {job.scheduler_id}")
         rv = self.get_condor().cancel_job(job_id=job.scheduler_id)
         logging.info(rv)
         logging.debug(f"{rv}")
 
+        cancel_job_msg2 = {"job_id": str(job_id), "requested_condor_deletion": 1}
```
jayrbolton: Can you make this `True` instead of `1` so it uses the right type?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -145,6 +146,13 @@ def _init_job_rec(self, user_id, params):
         with self.get_mongo_util().mongo_engine_connection():
             job.save()
 
+        create_job_status_msg = {
```
jayrbolton: I would have a field that goes in every message that is something like "event_type", which could be "created", "canceled", "queued", etc. That way consumers can switch on this field and know it'll always be there. It would also convey the meaning of each message/event.

lib/execution_engine2/utils/KafkaUtils.py
```diff
@@ -0,0 +1,61 @@
...
+import json
+import logging
+import sys
+
+from confluent_kafka import Producer, KafkaError, Consumer
+
+_TOPICS = ["ee2"]
+
+
+def _test_sample_consumer():
```
jayrbolton: I would put this in a test helpers module in a separate "test" package rather than having it in here.

lib/execution_engine2/utils/KafkaUtils.py
```diff
@@ -0,0 +1,61 @@
...
+
+def _delivery_report(err, msg):
+    if err is not None:
+        print("Message delivery failed:", err)
+    else:
+        print(f"Message delivered to topic '{msg.topic()}': {msg.value()}")
+
+
+def send_message_to_kafka(data, topic="ee2"):
+    required_keys = ["job_id", "new_status", "previous_status"]
```
jayrbolton: I would create a dataclass or a namedtuple if you want to safeguard yourself against errors. This way the type check would happen before runtime, or at the point when the data is created.

lib/execution_engine2/utils/KafkaUtilsLogger.py
jayrbolton: Should it be removed from the PR?

requirements.txt
```diff
@@ -31,3 +31,5 @@ codecov==2.0.15
 python-dotenv==0.10.3
 pre-commit==1.18.0
 python-dateutil==2.8.0
+confluent-kafka==1.2.0
+kafka==1.3.5
```
jayrbolton: Do you need both packages?

test/ee2_scheduler_kafka_test.py
```diff
@@ -0,0 +1,63 @@
...
+    send_message_to_kafka,
+    _test_sample_consumer,
+)
+import unittest
+
+
+class ExecutionEngine2SchedulerTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """
```
jayrbolton: You could remove the setUpClass method and put this docstring at the top of the class

test/ee2_scheduler_kafka_test.py
```diff
@@ -0,0 +1,63 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        pass
+
+    def test_kafka_producer(self):
+        message = {
+            "heading": "Dear Dolly",
+            "body": "Thank you for being a great cloned sheep.",
+        }
+        send_message_to_kafka(data=message)
```
jayrbolton: Can you add assertions here? One way to do it is to produce, then run a test consumer that consumes until you get the message you are expecting, with a max timeout of something like 180seconds

lib/execution_engine2/utils/KafkaUtils.py
```diff
@@ -0,0 +1,145 @@
...
+from dataclasses import dataclass
+
+from confluent_kafka import Producer
+
+KAFKA_EVENT_TYPE = "job_status_update"
+CONDOR_EVENT_TYPE = "condor_request"
+DEFAULT_TOPIC = "ee2"
+
+
+def send_kafka_update_status(data):
```
jayrbolton: Why have these five separate methods that don't seem to do anything? You could just call something like `send_message_to_kafka(msg)` where `msg` is an instance of a dataclass

lib/execution_engine2/utils/KafkaUtils.py
```diff
@@ -0,0 +1,145 @@
...
+    job_id: str = None
+    previous_status: str = None
+    new_status: str = None
+    scheduler_id: int = None
+    topic: str = DEFAULT_TOPIC
+    event_type: str = KAFKA_EVENT_TYPE
+    error: bool = False
+
+    def __post_init__(self):
+        if self.job_id is None:
```
jayrbolton: You don't need this manual type check if you take out the `= None` above in the field definition. So if you just have `job_id: str`, and you don't pass in a job_id when instantiating the object, then an error will get raised.

lib/execution_engine2/utils/KafkaUtils.py
```diff
@@ -0,0 +1,145 @@
...
+        if self.new_status is None:
+            raise Exception("Need to provide a new_status")
+
+        # All state transitions need a previous status, except the created one
+        if self.previous_status is None:
+            if self.new_status != Status.created.value:
+                raise Exception("Need to provide a previous_status")
+
+        # A created job may not have been able to have been submitted to condor, so it may not have a scheduler_id
+        if self.new_status not in ["created", "terminated", "canceled"]:
```
jayrbolton: Alternatively this can be an enum, see here: https://docs.python.org/3/library/enum.html
Then in the field definition you can make the type `new_status: StatusEnum` instead of str

lib/execution_engine2/utils/KafkaUtils.py
```diff
@@ -0,0 +1,145 @@
...
+                raise Exception("Need to provide error_code for unsuccessful jobs")
+
+
+def _delivery_report(err, msg):
+    if err is not None:
+        msg = "Message delivery failed:", err
+        logging.error(msg)
+
+
+def send_message_to_kafka(
```
jayrbolton: I would have this function take just the object instance and not the class

test/ee2_kafka_test.py
```diff
@@ -0,0 +1,114 @@
...
+            "new_status": "estimating",
+        }
+        with self.assertRaisesRegex(
+            expected_exception=Exception,
+            expected_regex="You must pass a scheduler id once the job has been created already.",
+        ):
+            send_kafka_update_status(data)
+
+        data["scheduler_id"] = "123"
+        send_kafka_update_status(data)
```
jayrbolton: Alternatively you can do:
```
data = KafkaStatusUpdate(
  job_id="123",
  previous_status="created",
  "new_status"="estimating"
)
send_message_to_kafka(data)
```
Having separate functions for each class and sending the class in the function is unintuitive to me

lib/execution_engine2/utils/KafkaUtils.py
bio-boris: @jayrbolton If I do that, I cannot inherit the dataclasses

Makefile
```diff
@@ -71,7 +71,8 @@ test:
 	nosetests -x -v --nocapture --nologcapture --with-coverage --cover-html --cover-package=execution_engine2 test/ee2_SDKMethodRunner_test.py
 	nosetests -x -v --nocapture --nologcapture --with-coverage --cover-html --cover-package=execution_engine2 test/ee2_MongoUtil_test.py
 	nosetests -x -v --nocapture --nologcapture --with-coverage --cover-html --cover-package=execution_engine2 test/ee2_server_test.py
-	nosetests -x -v --nocapture --nologcapture --with-coverage --cover-html --cover-package=execution_engine2 test/ee2_load_test.py
+	#nosetests -x -v --nocapture --nologcapture --with-coverage --cover-html --cover-package=execution_engine2 test/ee2_load_test.py
```
Tianhao-Gu: we should let travis run load test. it's the only integration test we have for some endpoints. if kafka code breaks that test, it's worth to figure it out.

bio-boris: I added it back in, see build https://travis-ci.org/kbase/execution_engine2/builds/627475530

PR #131, opened 2019-11-19
Add search to check jobs via mongo indexes and text_search

bio-boris:


bio-boris:
We should be really sure we want this or a bunch of indexes will be created.

test/ee2_SDKMethodRunner_test.py
```diff
@@ -1461,3 +1471,62 @@ def test_check_jobs_date_range(self, condor_mock):
...
+            creation_start_time=str(yesterday),
+            creation_end_time=str(tomorrow),
+            user=user_name,
+            freetext_search="weak_weak_weak",
+        )
+        print(job_states)
+        self.assertEquals(job_states["jobs"][0]["_id"], job_id1)
+
+        # So "weak" is not a match, but "weak_weak_weak" is a match
+        # So maybe we are better off with doing a filter on the jobs via iterating over the job_states object and if so, keep the record, if not, ignore it
```
Tianhao-Gu: I agree. I think we should just iterate over jobs. Otherwise, I am sure it will report back to us.

Issue #81, opened 2019-10-23
terminated, errored, and finished jobs sometimes don't have a finished time


eapearson:
In CI right now I'm seeing terminated jobs without a `finished` timestamp. I can work around this for the time being by using the `updated` timestamp.

```
{
	"_id": "5db07798f10c1461b1451daf",
	"user": "tgu2",
	"authstrat": "kbaseworkspace",
	"wsid": 28327,
	"status": "terminated",
	"updated": "2019-10-23 16:37:52.996000",
	"running": "2019-10-23 15:54:14.632000",
	"terminated_code": 0,
	"job_input": {
		"wsid": 28327,
		"method": "kb_uploadmethods.import_reads_from_staging",
		"params": [{
			"workspace_name": "tgu2:narrative_1516307095306",
			"import_type": "FASTQ/FASTA",
			"fastq_fwd_staging_file_name": "small.forward.fq",
			"fastq_rev_staging_file_name": null,
			"sra_staging_file_name": null,
			"sequencing_tech": "Illumina",
			"name": "small.forward.fq_reads",
			"single_genome": 1,
			"interleaved": 0,
			"insert_size_mean": null,
			"insert_size_std_dev": null,
			"read_orientation_outward": 0
		}],
		"service_ver": "b1fe0d58d1e722bc8e2fd991498b208cec1e7c00",
		"app_id": "kb_uploadmethods/import_fastq_sra_as_reads_from_staging",
		"source_ws_objects": [],
		"parent_job_id": "None",
		"narrative_cell_info": {
			"run_id": "308880f6-4538-42a8-a13c-5e8e475f38d6",
			"token_id": "975f55ac-1153-4806-9465-dbbd37a72483",
			"tag": "release",
			"cell_id": "996bf87c-b045-4f9d-af2a-0ba72a22a7ec"
		}
	},
	"job_id": "5db07798f10c1461b1451daf",
	"created": "2019-10-23 15:54:00+00:00",
	"app": {
		"id": "kb_uploadmethods/import_fastq_sra_as_reads_from_staging",
		"module_name": "kb_uploadmethods",
		"function_name": "import_fastq_sra_as_reads_from_staging",
		"title": "Import FASTQ/SRA File as Reads from Staging Area"
	}
}
```

eapearson:
I've noticed this for an error as well (some error messages, not all).
I don't think this is intentional since most errored jobs do have a finished time.

eapearson:
[...]

PR #132, opened 2019-11-19
#81 verify jobs

bio-boris:
#81 verify jobs

### DIFF
--- a/lib/execution_engine2/db/models/transfer/check_jobs_for_finish_times.py
+++ b/lib/execution_engine2/db/models/transfer/check_jobs_for_finish_times.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+try:
+    from .transfer_ujs_njs import MigrateDatabases
+    from execution_engine2.db.models.models import Status, valid_status
+except:
+    from transfer_ujs_njs import MigrateDatabases
+    from models import *
+
+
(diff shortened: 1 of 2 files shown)


bio-boris:
I ran this script against CI and it didn't throw any errors, so I think this issue is fixed. Unless my script needs some revisions.

lib/execution_engine2/db/models/transfer/check_jobs_for_finish_times.py
```diff
@@ -0,0 +1,40 @@
...
+        valid_status(status)
+    except:
+        raise Exception(job_id)
+    if count % 500 == 0:
+        print(f"Processed {count} records")
+
+    if status in [Status.error.value, Status.completed.value, Status.terminated.value]:
+        end_time = job.get("finished")
+        if end_time is None:
+            raise Exception("End time is NONE", job_id)
```
Tianhao-Gu: do we want to raise here or return all bad job ids and fix them in a batch?

bio-boris: I think raise is good so the script crashes and I can just start looking at the database. That's what I did already  and fixed the jobs below

Tianhao-Gu: 👍

Issue #136, opened 2019-11-20
check_job and check_jobs - queued timestamp is float not integer


eapearson:
In both the check_job and check_jobs methods, the timestamp `queued` is provided as a float seconds rather than integer milliseconds. 

For example, in the request 

```json
{
  "version": "1.1",
  "id": "123",
  "method": "execution_engine2.check_job",
  "params": [
    {
      "job_id": "5dd540027ec5a1da6ebb0f09"
    }
  ]
}
```

with the result

```json
{
    "version": "1.1",
    "result": [
        {
            "user": "eaptest30",
            "authstrat": "kbaseworkspace",
            "wsid": 42900,
            "status": "terminated",
            "updated": 1574256658553,
            "queued": 1574256643.672457,
            "finished": 1574256658553,
            "terminated_code": 0,
            "scheduler_type": "condor",
            "scheduler_id": "12900",
            "job_input": {
                "wsid": 42900,
                "method": "kb_SPAdes.run_SPAdes",
                "params": [
                    {
                        "workspace_name": "eaptest30:narrative_1559318949611",
                        "read_libraries": [
                            "42900/3/4"
                        ],
                        "output_contigset_name": "SPAdes.contigs",
                        "dna_source": "standard",
                        "min_contig_length": 500,
                        "kmer_sizes": [],
                        "skip_error_correction": 0
                    }
                ],
                "service_ver": "e0a0f9df50bb6765db40f6949b18c7a016610801",
                "app_id": "kb_SPAdes/run_SPAdes",
                "source_ws_objects": [
                    "42900/3/4"
                ],
                "parent_job_id": "None",
                "narrative_cell_info": {
                    "run_id": "8d5ad754-d29d-4043-9439-29f5f8b536d3",
                    "token_id": "e5f189ce-3c4a-4073-8cf9-e1e37db5a926",
                    "tag": "release",
                    "cell_id": "9565150f-868e-4627-89c2-65de4f8b0b37"
                }
            },
            "job_id": "5dd540027ec5a1da6ebb0f09",
            "created": 1574256642000
        }
    ],
    "id": "123"
}
```

Tianhao-Gu:
@eapearson Thanks. It should be fixed now.

eapearson:
Thanks, looks good now.

PR #137, opened 2019-11-20
check_jobs queued_time

Tianhao-Gu:


Tianhao-Gu:
addressing https://github.com/kbase/execution_engine2/issues/136

Issue #133, opened 2019-11-19
Iron out ee2_admin vs ee2_admin_RO 


bio-boris:
ee2_admin can 
(via the admin function)
* Cancel_jobs
* Modify jobs in the future
* Qedit jobs in the future
(via the check_jobs_functions and view_logs functions)
* view_job_logs
* view_job_records

ee2_read_only_admin
(via the check_jobs_functions and view_logs functions)

* view_job_logs for other users
* view_job_records for other users

the is_admin endpoint can return a T/F

Should we also have endpoints
can_read_ee2()
can_write_ee2()?

@Tianhao-Gu @briehl @eapearson @scanon

bio-boris:
Solved Here https://github.com/kbase/execution_engine2/pull/189

PR #149, opened 2019-11-28
#133 Added endpoint to check admin privs

bio-boris:
See #133

### DIFF
--- a/lib/execution_engine2/SDKMethodRunner.py
+++ b/lib/execution_engine2/SDKMethodRunner.py
@@ -355,7 +355,7 @@ class SDKMethodRunner:
         self.workspace = None
         self.workspace_auth = None
 
-        self.admin_roles = config.get("admin_roles", ["EE2_ADMIN"])
+        self.admin_roles = config.get("admin_roles", ["EE2_ADMIN", "EE2_ADMIN_RO"])
 
         catalog_url = config.get("catalog-url")
         self.catalog = Catalog(catalog_url)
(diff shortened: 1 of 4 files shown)


lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -960,9 +972,26 @@ def get_sort_order(ascending):
...
+    def get_admin_permission(self):
+        """
+        Get your your type of admin permissions
+        :return:
+        """
+        aau = AdminAuthUtil(self.auth_url, self.admin_roles)
+        roles = list(aau._fetch_user_roles(self.token))
+        permission = None
+        if "EE2_ADMIN" in roles:
+            permission = "w"
```
Tianhao-Gu: ok. But spec says permission is `One of ('read' | 'write' | 'none')`. Just in case people will complain, we should update either one.

lib/execution_engine2/execution_engine2Server.py
```diff
@@ -12,8 +12,13 @@
 from wsgiref.simple_server import make_server
 
 import requests as _requests
-from jsonrpcbase import JSONRPCService, InvalidParamsError, KeywordError, \
```
Tianhao-Gu: I didn't read through changes in this file. I assume it's an auto linter does all the changes. It shouldn't probably turn off on sdk auto-generated files.

lib/execution_engine2/SDKMethodRunner.py
bio-boris: Fixed, thanks

lib/execution_engine2/execution_engine2Server.py
bio-boris: Right now it just skips the impl file

https://github.com/kbase/execution_engine2/blob/develop/.pre-commit-config.yaml

PR #153, opened 2019-12-10
add job_log_exists

Tianhao-Gu:


lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -312,16 +316,18 @@ def add_job_logs(self, job_id, log_lines):
         :return:
         """
         logging.debug(f"About to add logs for {job_id}")
-        job = self.get_mongo_util().get_job(job_id=job_id)
-        self._test_job_permissions(job, job_id, JobPermissions.WRITE)
-        logging.debug("Success, you have permission to add logs for " + job_id)
+        mongo_util = self.get_mongo_util()
+        # job = mongo_util.get_job(job_id=job_id)
+        # self._test_job_permissions(job, job_id, JobPermissions.WRITE)
+        # logging.debug("Success, you have permission to add logs for " + job_id)
```
Tianhao-Gu: @bio-boris Do we need to test job permission here? If I remove those, locally, adding 200 job logs takes 22 seconds vs 210 seconds if I keep them.

Tianhao-Gu:
@bio-boris it reduces run time from 300" to less than 100" locally adding 200 job logs.

PR #154, opened 2019-12-11
add job_log_exists (#153)

bio-boris:
* add job_log_exists

* using pymongo to fetch job logs

* remove unused comments

* add more tests

* remove check permission

* fetch job

* reduce thread due to travis time limit

### DIFF
--- a/lib/execution_engine2/SDKMethodRunner.py
+++ b/lib/execution_engine2/SDKMethodRunner.py
@@ -224,8 +224,7 @@ class SDKMethodRunner:
         :return:
         """
 
-        mongo_util = self.get_mongo_util()
-        log = mongo_util.get_job_log_pymongo(job_id)
+        log = self.get_mongo_util().get_job_log_pymongo(job_id)
 
         lines = []
         for log_line in log.get("lines", []):  # type: LogLines
--- a/lib/execution_engine2/SDKMethodRunner.py
+++ b/lib/execution_engine2/SDKMethodRunner.py
@@ -228,8 +228,9 @@ class SDKMethodRunner:
         :return:
         """
         logging.debug(f"About to view logs for {job_id}")
-        job = self.get_mongo_util().get_job(job_id)
-        self._test_job_permissions(job, job_id, level=JobPermissions.READ)
+        if not self._get_job_permission_from_cache(job_id, self.user_id, JobPermissions.READ):
+            job = self.get_mongo_util().get_job(job_id)
+            self._test_job_permissions(job, job_id, level=JobPermissions.READ)
         logging.debug("Success, you have permission to view logs for " + job_id)
         return self._get_job_log(job_id, skip_lines)
 
@@ -411,7 +412,8 @@ class SDKMethodRunner:
         # Maybe if the call fails, we don't actually cancel the job?
         logging.debug(f"Attempting to cancel job {job_id}")
         job = self.get_mongo_util().get_job(job_id=job_id)
-        self._test_job_permissions(job, job_id, JobPermissions.WRITE)
+        if not self._get_job_permission_from_cache(job_id, self.user_id, JobPermissions.WRITE):
+            self._test_job_permissions(job, job_id, JobPermissions.WRITE)
         logging.info(f"User has permission to cancel job {job_id}")
         logging.debug(f"User has permission to cancel job {job_id}")
         self.get_mongo_util().cancel_job(job_id=job_id, terminated_code=terminated_code)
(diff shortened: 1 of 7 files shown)


Tianhao-Gu:
@bio-boris I think performance for add_job_logs and check_job_logs are acceptable now. Do you mind taking a look at this PR?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -194,21 +199,25 @@ def _get_job_log(self, job_id, skip_lines):
         :param skip_lines:
         :return:
         """
-        log = self.get_mongo_util().get_job_log(job_id)
+
+        mongo_util = self.get_mongo_util()
+        with mongo_util.me_collection(self.config["mongo-logs-collection"]):
```
bio-boris: do we need this "with" statement here? Isn't there one inside the `get_job_log_pymongo` function? If we need it, can we move this with statement into that function?

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -312,16 +322,26 @@ def add_job_logs(self, job_id, log_lines):
...
+            self._test_job_permissions(job, job_id, JobPermissions.WRITE)
+
         logging.debug("Success, you have permission to add logs for " + job_id)
 
-        try:
-            log = self.get_mongo_util().get_job_log(job_id=job_id)
-        except RecordNotFoundException:
-            log = self._create_new_log(pk=job_id)
+        with mongo_util.me_collection(self.config["mongo-logs-collection"]):
+            try:
```
bio-boris: can we move this with into the get_job_log_pymongo function?

Tianhao-Gu: 👍

lib/execution_engine2/SDKMethodRunner.py
Tianhao-Gu: 👍

Tianhao-Gu:
@bio-boris thanks. moved `with` clause to function. it's still taking about 30 seconds to add 200 job logs in CI. runs really fast locally though. hmmm.

lib/execution_engine2/SDKMethodRunner.py
```diff
@@ -657,7 +710,8 @@ def get_job_status(self, job_id):
             raise ValueError("Please provide valid job_id")
 
         job = self.get_mongo_util().get_job(job_id=job_id)
-        self._test_job_permissions(job, job_id, JobPermissions.READ)
+        if not self._get_job_permission_from_cache(job_id, self.user_id, JobPermissions.READ):
```
bio-boris: What do you think about moving `._get_job_permission_from_cache` into  `_test_job_permissions`

Tianhao-Gu: the problem moving it to `_test_job_permissions` is we cannot limit calls of `self.get_mongo_util().get_job` which is the bottleneck that slows things down. If we move `get_job` into `_test_job_permissions` as well, there are other functions like `cancel_job` does need job after `_test_job_permissions`

bio-boris: Ok that makes sense.
What do you think about about moving these 3 lines into a new function called get_job_and_permission or something like that? It can return the job object if necessary, but it hides that you have to make the calls to the cache or not.

Tianhao-Gu:
@bio-boris Do you mind taking another look?

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -164,6 +190,10 @@ def get_job_log(self, job_id: str = None) -> JobLog:
 
         return job_log
 
+    def job_log_exists(self, job_id):
```
bio-boris: is this used somewhere?

Tianhao-Gu: no want me to remove it? thought it might be useful somewhere down the road.

Tianhao-Gu:
@bio-boris removed unused `job_log_exists`

lib/execution_engine2/db/MongoUtil.py
Tianhao-Gu: removed.

bio-boris:
@Tianhao-Gu Looks good we can merge this in.
Two questions/ideas
* Should _get_job_with_permission() be renamed to _get_job()?
* does the context manager open up two connections to mongo? Do we need both? Maybe it would be faster to return only the pymongo connection instead of both the pymongo and the me connection. I don't actually know though

Tianhao-Gu:
@bio-boris 👍 updated context manager to use only pymongo connection.

PR #164, opened 2019-12-16
rollback jobs

Tianhao-Gu:


Tianhao-Gu:
@bio-boris Do you mind taking a look at it first?

lib/execution_engine2/db/models/transfer/rollback_db.py
```diff
@@ -0,0 +1,238 @@
...
+
+        ujs_job_doc = {"_id": job_id,
+                       "user": ee2_job.get("user"),
+                       "authstrat": ee2_job.get("authstrat"),
+                       "authparam": ee2_job.get("authstrat"),
+                       "created": job_id.generation_time,
+                       "updated": datetime.fromtimestamp(ee2_job.get("updated")),
+                       "estcompl": None,
+                       "service": ee2_job.get("user"),
+                       "complete": ee2_job.get("status") in ["completed", "terminated", "error"],
```
bio-boris: it would be interesting to see how many jobs are not complete vs complete at the end of insertion.

Issue #108, opened 2019-11-12
get_job_logs should take offset and limit


eapearson:
Current the get_job_logs method takes the parameter "skip_lines" as the only control over the log entries returned.

We also need a limit on the number of entries returned, in order to support paging. This is commonly named "limit".

In fact, I'd recommend adding the more standard parameter "offset" along with its partner "limit", to support paging (using the parameter names used in other ee2 methods and other services.)

For compatibility, skip_lines could be retained. Not sure how to handle both being supplied, but probably an error would be best, rather than some precedence or mixing of them (e.g. limit and skip_lines)

For "offset" and "limit" I don't think we should allow limit to be omitted, because it is unclear whether the service would impose a default (e.g. 1000) is or whether the behavior would be "unlimited" as it is now. 

A user of the api can use offset with a high limit to achieve fetching all log entries, yet still retain some control and safety.

bio-boris:
> For "offset" and "limit" I don't think we should allow limit to be omitted, because it is unclear whether the service would impose a default (e.g. 1000) is or whether the behavior would be "unlimited" as it is now.


@eapearson The limit is 2000 right now
https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/SDKMethodRunner.py#L993-L995

eapearson:
@bio-boris on my side I'm requiring that the offset and limit always be specified, so I don't plan on using the default value. I do of course see the value of having a default value for casual usage, but on the ui side I anticipate that accessing logs will always be paged (even if the display side appears to be scrolling), and that requires always specifying the offset and limit.

PR #196, opened 2020-03-09
add offset and limit to get_job_logs

Tianhao-Gu:
add offset and limit to `get_job_logs`

https://github.com/kbase/execution_engine2/issues/108

### DIFF
--- a/execution_engine2.spec
+++ b/execution_engine2.spec
@@ -195,12 +195,15 @@
 
 
         /*
-            skip_lines - optional parameter, number of lines to skip (in case they were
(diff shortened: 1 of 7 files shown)


Tianhao-Gu:

bio-boris:
Looks good. We could probably use a test for throwing an exception for providing skip lines and offset at the same time.

bio-boris:
At some point we should probably slice up the array in mongo and not send the whole thing back and make the whole thing sit in memory.

Tianhao-Gu:
@bio-boris Thanks. Added test for an exception.

Issue #176, opened 2020-01-17
Service Wizard URL not being populated


bio-boris:
Service Wizard URL is not being populated

```
root@93b65355c19f:/kb/module# cat deploy.cfg
[simpleapp]
kbase-endpoint = https://ci.kbase.us/services
job-service-url =
workspace-url = https://ci.kbase.us/services/ws
shock-url = https://ci.kbase.us/services/shock-api
handle-service-url = https://ci.kbase.us/services/handle_service
srv-wiz-url =
njsw-url =
auth-service-url = https://ci.kbase.us/services/auth/api/legacy/KBase/Sessions/Login
auth-service-url-allow-insecure = false
```

bio-boris:
https://github.com/search?l=Python&p=1&q=self.srv_wiz_url&type=Code

bio-boris:
https://github.com/kbase/JobRunner/commit/fc26d02b166603bcf077c8b1b4e0d732a6afa749

https://github.com/kbase/execution_engine2/commit/5684952028baad8a6c8a54b8c1b0117f4009614b

PR #210, opened 2020-04-03
Srv wiz url was not in config

bio-boris:
https://github.com/kbase/execution_engine2/issues/176


bio-boris:
https://github.com/kbase/JobRunner/pull/19

PR #217, opened 2020-04-07
Held job reaper

bio-boris:
* Add held jobs endpoint 
* Add reaper processs for held jobs
* Copy requests into test.utils partially.
* move test from sdkmrtest into sdkr.status test ... more tests moved out in future


bin/PurgeHeldJobs.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python3
```
bio-boris, line 1: New script to process held jobs , to be run by start_server.sh

execution_engine2.spec
```diff
@@ -527,6 +527,13 @@
         funcdef check_jobs_date_range_for_user(CheckJobsDateRangeParams params) returns (CheckJobsResults) authentication required;
         funcdef check_jobs_date_range_for_all(CheckJobsDateRangeParams params) returns (CheckJobsResults) authentication required;
 
+        typedef structure {
+            UnspecifiedObject held_job;
+        } HeldJob;
+        /*
+            Handle a held CONDOR job. You probably never want to run this, only the reaper should run it.
+        */
+        funcdef handle_held_job(string cluster_id) returns (HeldJob) authentication required;
```
bio-boris, line 536: This returns a full job record right now.. Dunno if I should make it a specific type..

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -26,9 +26,15 @@ class JobPermissions(Enum):
     NONE = "n"
 
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
bio-boris, line 31: This is how to avoid circular imports

PR #255, opened 2020-09-23
DataUp-181 Prevent Reaper from quitting

bio-boris:

Prevent the job reaper from exiting on error, just continue trying to reap

# Testing Instructions
Didn't write tests for this script.

# Dev Checklist:

- [ ] My code follows the guidelines at https://sites.google.com/truss.works/kbasetruss/development
- [ ] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the travis build passes)

### DIFF
--- a/bin/PurgeHeldJobs.py
+++ b/bin/PurgeHeldJobs.py
@@ -173,4 +173,4 @@ if __name__ == "__main__":
             time.sleep(5)
         except Exception as e:
             slack_client.ee2_reaper_failure(endpoint=ee2_endpoint)
-            sys.exit(f"{e}")
+
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # execution_engine2 (ee2) release notes
 =========================================
+## 0.0.3.4
+  * Change 7 day periodic_remove to 7 day hold
+  * Prevent reaper from prematurely exiting
+  
 ## 0.0.3.3
   * Change log appends from update to $push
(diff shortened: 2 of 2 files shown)


bio-boris:
Should try catch the slack client call as well!!

bin/PurgeHeldJobs.py
```diff
@@ -173,4 +173,4 @@ def handle_hold_event(event):
             time.sleep(5)
         except Exception as e:
             slack_client.ee2_reaper_failure(endpoint=ee2_endpoint)
-            sys.exit(f"{e}")
```
leia-sefkin, line 176: is there any concern over keeping the reaper running when there's a legitimate error here?

leia-sefkin (commented):
I'm a bit out of context here so not sure my review can be all that helpful. Seems like an ok thing to try, but I'm not sure of the consequences of not quitting when we do encounter and exception here. Other concerns might be whether slack is the best location to be recording these errors - if that's a reliable process.

eamahanna (commented):
My thoughts largely align with Leia's. I'm not sure the reason and consequences for this change?

bin/PurgeHeldJobs.py
bio-boris: No, the reaper should never end

bio-boris:
> My thoughts largely align with Leia's. I'm not sure the reason and consequences for this change?

The reason is because I saw the reaper had exited, when it should have kept running

Tianhao-Gu (approved):
👍 GLTM

PR #271, opened 2020-12-04
DATAUP-259 Check child jobs

bio-boris:

* Get the status of a job and the status of its children
* Tests are running

# Testing Instructions
- [x] Tests pass in travis and locally


execution_engine2.spec
```diff
@@ -391,8 +391,22 @@
         funcdef check_job(CheckJobParams params) returns (JobState job_state) authentication required;
 
         /*
-            job_states - states of jobs
+            parent_job - state of parent job
+            job_states - states of child jobs
+            IDEA: ADD aggregate_states - count of all available child job states, even if they are zero
```
ialarmedalien, line 396: should this line be in here?

execution_engine2.spec
```diff
@@ -391,8 +391,22 @@
...
+        */
+        typedef structure {
+            JobState parent_jobstate;
+            list<JobState> child_jobstates;
+        } CheckJobBatchResults;
+
+        /*
+            get current status of a parent job, and it's children, if it has any.
```
ialarmedalien, line 404: Punctuation police alert:
`it's` = it is
`its` = belonging to it

bio-boris: ok fixing it in next pr

execution_engine2.spec
bio-boris: Fixing this in next PR

lib/execution_engine2/sdk/SDKMethodRunner.py
```diff
@@ -319,6 +317,41 @@ def get_job_status_field(self, job_id, as_admin=False):
         """ Authorization Required: Read """
         return self.get_jobs_status().get_job_status(job_id=job_id, as_admin=as_admin)
 
+    def check_job_batch(
+        self,
+        parent_job_id,
+        check_permission=True,
```
ialarmedalien, line 323: Does `check_job_batch` get called elsewhere, other than in Impl file? The implementation file doesn't have any way of adding the `check_permission` param.

lib/execution_engine2/sdk/SDKMethodRunner.py
```diff
@@ -319,6 +317,41 @@ def get_job_status_field(self, job_id, as_admin=False):
...
+        if child_job_ids:
+            child_job_states = self.get_jobs_status().check_jobs(
+                job_ids=child_job_ids,
+                check_permission=True,
+                exclude_fields=exclude_fields,
+                return_list=1,
+            )["job_states"]
+        return {
+            "parent_jobstate": parent_job_status,
+            "child_jobstates": child_job_states,
```
ialarmedalien, line 352: is this going to be an array of all job states, or will it be a mapping with job ID => job state?

bio-boris: array of all job states

lib/execution_engine2/sdk/SDKMethodRunner.py
bio-boris: No. This is simply a copy of the way things are done for "check_job". We want the path from the IMPL to always check permissions. Do you want me to move it into the function definition instead?

lib/execution_engine2/sdk/SDKMethodRunner.py
ialarmedalien: So how are you going to know which job is in which state?

lib/execution_engine2/sdk/SDKMethodRunner.py
ialarmedalien: Easiest would probably be to add the param to the function call in the Impl file.

lib/execution_engine2/sdk/SDKMethodRunner.py
bio-boris: @ialarmedalien It is a list of job states, not a list of job statuses. You can check the job[status] field to see the job status.

lib/execution_engine2/sdk/SDKMethodRunner.py
bio-boris: @ialarmedalien We don't want the user to be able to bypass permissions, but we've used it in the past before we had the permissions request mocking and before we had the permissions cache.

lib/execution_engine2/sdk/SDKMethodRunner.py
ialarmedalien: Ah sorry, gotcha!

ialarmedalien (approved):
👍

PR #296, opened 2021-01-28
pass a UserClientSet into SDKMethodRunner

MrCreosote:

UCS contains the clients that need to be instantiated on a per user basis.
UCS is passed in to SDKMR to 1) simplify the SDKMR constructor, and 2) allow for eventual unit testing of SDKMR, since the UCS and workspace client can be mocked via create_autospec.

Also add instructions for running pre-commit hooks.

# Testing Instructions
- [x] Tests pass in travis and locally 

# Dev Checklist:

- [?] My changes generate no new warnings - too many warnings to feasibly check
- [x] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the travis build passes)

### DIFF
--- a/lib/execution_engine2/constants.py
+++ b/lib/execution_engine2/constants.py
@@ -1,0 +1,5 @@
+'''
+Global EE2 constants.
+'''
+ADMIN_READ_ROLE = "EE2_ADMIN_RO"
+ADMIN_WRITE_ROLE = "EE2_ADMIN"
--- a/lib/execution_engine2/constants.py
+++ b/lib/execution_engine2/constants.py
@@ -1,5 +1,3 @@
-"""
-Global EE2 constants.
-"""
+""" Global EE2 constants. """
 ADMIN_READ_ROLE = "EE2_ADMIN_RO"
 ADMIN_WRITE_ROLE = "EE2_ADMIN"
--- a/lib/execution_engine2/constants.py
+++ b/lib/execution_engine2/constants.py
@@ -1,5 +1,5 @@
-'''
+"""
 Global EE2 constants.
-'''
+"""
 ADMIN_READ_ROLE = "EE2_ADMIN_RO"
 ADMIN_WRITE_ROLE = "EE2_ADMIN"
--- a/lib/execution_engine2/constants.py
+++ b/lib/execution_engine2/constants.py
@@ -1,3 +1,0 @@
-""" Global EE2 constants. """
-ADMIN_READ_ROLE = "EE2_ADMIN_RO"
-ADMIN_WRITE_ROLE = "EE2_ADMIN"
--- a/lib/execution_engine2/execution_engine2Impl.py
+++ b/lib/execution_engine2/execution_engine2Impl.py
@@ -6,7 +6,7 @@ from cachetools import TTLCache
 
 from lib.execution_engine2.db.MongoUtil import MongoUtil
 from lib.execution_engine2.sdk.SDKMethodRunner import SDKMethodRunner
-from execution_engine2.utils.clients import get_user_client_set, UserClientSet
+from execution_engine2.utils.APIHelpers import GenerateFromConfig
 
 
 #END_HEADER
@@ -42,10 +42,6 @@ class execution_engine2:
 
     ADMIN_ROLES_CACHE_SIZE = 500
     ADMIN_ROLES_CACHE_EXPIRE_TIME = 300  # seconds
-
-    def get_user_clients(self, ctx) -> UserClientSet:
-        return get_user_client_set(self.config, ctx['user_id'], ctx['token'])
-
     #END_CLASS_HEADER
 
     # config contains contents of config file in a hash or None if it couldn't
@@ -64,9 +60,7 @@ class execution_engine2:
             maxsize=self.ADMIN_ROLES_CACHE_SIZE, ttl=self.ADMIN_ROLES_CACHE_EXPIRE_TIME
         )
         self.mongo_util = MongoUtil(config)
-
-
-
+        self.gen_cfg = GenerateFromConfig(config)
         #END_CONSTRUCTOR
         pass
 
@@ -233,7 +227,7 @@ class execution_engine2:
         #BEGIN run_job
         mr = SDKMethodRunner(
             self.config,
-            user_clients=self.get_user_clients(ctx),
+            user_clients=self.gen_cfg.get_user_clients(ctx),
             job_permission_cache=self.job_permission_cache,
             admin_permissions_cache=self.admin_permissions_cache, mongo_util=self.mongo_util
         )
@@ -305,7 +299,7 @@ class execution_engine2:
         #BEGIN run_job_batch
         mr = SDKMethodRunner(
             self.config,
-            user_clients=self.get_user_clients(ctx),
+            user_clients=self.gen_cfg.get_user_clients(ctx),
             job_permission_cache=self.job_permission_cache,
             admin_permissions_cache=self.admin_permissions_cache, mongo_util=self.mongo_util
         )
@@ -334,7 +328,7 @@ class execution_engine2:
         #BEGIN abandon_children
         mr = SDKMethodRunner(
             self.config,
-            user_clients=self.get_user_clients(ctx),
+            user_clients=self.gen_cfg.get_user_clients(ctx),
             job_permission_cache=self.job_permission_cache,
             admin_permissions_cache=self.admin_permissions_cache, mongo_util=self.mongo_util
         )
@@ -417,7 +411,7 @@ class execution_engine2:
         #BEGIN run_job_concierge
         mr = SDKMethodRunner(
             self.config,
-            user_clients=self.get_user_clients(ctx),
+            user_clients=self.gen_cfg.get_user_clients(ctx),
             mongo_util=self.mongo_util
         )
         job_id = mr.run_job_concierge(params=params,concierge_params=concierge_params)
@@ -487,7 +481,7 @@ class execution_engine2:
         #BEGIN get_job_params
         mr = SDKMethodRunner(
             self.config,
-            user_clients=self.get_user_clients(ctx),
+            user_clients=self.gen_cfg.get_user_clients(ctx),
             job_permission_cache=self.job_permission_cache,
             admin_permissions_cache=self.admin_permissions_cache,
             mongo_util=self.mongo_util
@@ -516,7 +510,7 @@ class execution_engine2:
         #BEGIN update_job_status
         mr = SDKMethodRunner(
             self.config,
-            user_clients=self.get_user_clients(ctx),
+            user_clients=self.gen_cfg.get_user_clients(ctx),
             job_permission_cache=self.job_permission_cache,
             admin_permissions_cache=self.admin_permissions_cache,
             mongo_util=self.mongo_util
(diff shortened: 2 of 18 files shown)


MrCreosote:
Huh. The failing tests were marked as failing immediately, and failing after [2, 8]m, even thought they couldn't have started yet

lib/execution_engine2/constants.py
```diff
@@ -0,0 +1,5 @@
+'''
```
MrCreosote, line 1: I had to pull the constants into a separate file to avoid an import cycle

lib/execution_engine2/utils/clients.py
```diff
@@ -0,0 +1,111 @@
...
+        return self._token
+
+    def workspace(self):
+        return self._workspace
+
+    def workspace_auth(self):
+        return self._workspace_auth
+
+
+class ClientSet:
```
MrCreosote, line 76: Currently unused, but is the next step in the unit-testability work

lib/execution_engine2/constants.py
```diff
@@ -0,0 +1,3 @@
+""" Global EE2 constants. """
+ADMIN_READ_ROLE = "EE2_ADMIN_RO"
```
bio-boris, line 2: We have a EE2Constants.py file already. Maybe we should use that?

MrCreosote: Good call, fixed

MrCreosote:
I see a fix I can make for the workspace-url check in the client set, give me a sec

MrCreosote:
@bio-boris ok, all set, ready for re-review

lib/execution_engine2/execution_engine2Impl.py
```diff
@@ -41,6 +42,10 @@ class execution_engine2:
 
     ADMIN_ROLES_CACHE_SIZE = 500
     ADMIN_ROLES_CACHE_EXPIRE_TIME = 300  # seconds
+
+    def get_user_clients(self, ctx) -> UserClientSet:
```
bio-boris, line 46: Is there an advantage of having this here instead of inside the SDKMethodRunner?

lib/execution_engine2/utils/clients.py
```diff
@@ -0,0 +1,112 @@
...
+        self.condor = Condor(cfg_path)
+        self.catalog_utils = CatalogUtils(cfg["catalog-url"], cfg["catalog-token"])
+        auth_url = cfg["auth-url"]
+        self.auth = KBaseAuth(auth_url=auth_url + "/api/legacy/KBase/Sessions/Login")
+        # TODO using hardcoded roles for now to avoid possible bugs with mismatched cfg roles
+        #       these should probably be configurable
+        self.auth_admin = AdminAuthUtil(auth_url, [ADMIN_READ_ROLE, ADMIN_WRITE_ROLE])
+
+        # KafkaClient has a nice error message when the arg is None
+        self.kafka_client = KafkaClient(cfg.get("kafka-host"))
```
bio-boris, line 108: Maybe we should follow the fail fast with Keyerror way to handling dictionaries by doing the following
```
cfg['key'] == required value
cfg.get('key') == optionally None
cfg.get('key','default') == explicitly set default
```

lib/execution_engine2/utils/clients.py
```diff
@@ -0,0 +1,112 @@
...
+        # below will need lots of params so a dict makes sense there, maybe keep the apis similar?
+        # TODO the client throws a 'X is not a valid url' error if the url isn't valid, improve
+        #      by catching & rethrowing with a more clear message that the config is wrong
+        ws_url = cfg.get("workspace-url")  # may want to make the keys constants?
+        if not ws_url or not ws_url.strip():
+            raise ValueError("missing workspace-url in configuration")
+        self._workspace = Workspace(ws_url, token=token)
+        self._workspace_auth = WorkspaceAuth(user_id, self._workspace)
+
+    # create_autospec can't mock instance variables, so we add some boilerplate
```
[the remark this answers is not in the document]
MrCreosote: Have you verified that with a `@patch` decorator with `spec_set=True` that it detects instance variables? My understanding is that the `spec_set` mocks can't see instance variables regardless of how the mock is created. I'll update the comment to that effect.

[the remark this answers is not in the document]
bio-boris: Ah, No, instance variables are not detected at all, so nevermind.

lib/execution_engine2/utils/clients.py
```diff
@@ -0,0 +1,112 @@
...
+    # Could make these properties but then they return MagicMocks which don't update with API
+    # changes
+
+    def user_id(self):
```
bio-boris, line 63: This seems  strange to do it this way. Are other kbase codebases doing it this way?

test/tests_for_auth/ee2_admin_mode_test.py
```diff
@@ -104,13 +108,22 @@ def get_runner_with_condor(self) -> SDKMethodRunner:
...
+
     @patch.object(Catalog, "get_module_version", return_value="module.version")
-    @patch.object(WorkspaceAuth, "can_write", return_value=True)
     @patch.object(AdminAuthUtil, "_fetch_user_roles")
-    def test_regular_user(self, aau, workspace, catalog):
+    def test_regular_user(self, aau, catalog):
         # Regular User
         lowly_user = "Access Denied: You are not an administrator"
-        runner = self.getRunner()
+        ucs, wsa = self.get_mocks()
```
bio-boris, line 123: It would help with readability to write out 
`user_client_set` and `ws_auth` instead of these 3 letter variables names

test/tests_for_sdkmr/ee2_SDKMethodRunner_EE2Logs_test.py
```diff
@@ -35,7 +36,7 @@ def setUpClass(cls):
         cls.token = "token"
 
         cls.method_runner = SDKMethodRunner(
-            cls.cfg, user_id=cls.user_id, token=cls.token
+            cls.cfg, UserClientSet(cls.cfg, cls.user_id, cls.token)
```
bio-boris, line 39: The codebase usually does the following

def foo(a):
is called 
`foo(a)`

`def foo(a,b):`
is called
`foo(a=1,b=1)`

bio-boris: So we should probably have some consistency in the codebase.

lib/execution_engine2/execution_engine2Impl.py
MrCreosote: yes, instantiating the client set outside of SDKMR is what allows for passing mock objects into SDKMR rather than monkey patching (although you can still do that if you want) and enables constructor-based [dependency injection](https://www.freecodecamp.org/news/a-quick-intro-to-dependency-injection-what-it-is-and-when-to-use-it-7578c84fa88f/). The major change for this PR is exactly that - making the code that creates SDKMR responsible for the creation of SDKMR's dependencies rather than SDKMR.

lib/execution_engine2/utils/clients.py
MrCreosote: If we wanted to do something like this the way it'd do it is something like:
```
if not cfg.get('key') or not cfg.get('key').strip():
    raise ValueError('configuration is missing key "key"')
```
Just getting a KeyError("key") is not very helpful to someone trying to figure out why the service won't start, especially if they're not familiar with it.

I actually did use the first approach in cases where the client I was instantiating ignored missing arguments, although there could still be issues if the argument was whitespace or a bad url, etc. Where the clients had readable error messages (like Kafka and Slack) I went with those.

In most of my repos all the config args are checked for correctness, and urls used to make requests to the service and check the results, before the service starts. The checking is mostly delegated to the clients / objects that use them since they understand the semantics.

lib/execution_engine2/utils/clients.py
MrCreosote: I don't think so. You only need to do this when you have a mock where you need to mock the instance variables, which doesn't happen all that often in my experience. 

A better way to do this might be to make a constructor that takes a workspace client, user id, and token, and then you can just mock the workspace client and you don't need to mock the client set. I think I'll do that, then this becomes a non-issue.

test/tests_for_sdkmr/ee2_SDKMethodRunner_EE2Logs_test.py
MrCreosote: I thought that you used keyword arguments for optional arguments and positional arguments for required arguments? E.g. the cfg is required, so it's not a keyword in either version, and UCS is required in the new version so it's not a keyword arg either.

bio-boris: @MrCreosote If you look at the codebase, that is the pattern that is often used, so we should stick to one way and do it that way, in order to maintain consistency.

lib/execution_engine2/execution_engine2Impl.py
bio-boris: I should have said inside the SDKMethodRunner file, defined outside of the SDKMethodRunner class, or moving this function to its own file. We should probably move out everything from the BEGIN_CLASS_HEADER section to its own file, as there are often issues with editing code in the impl file, and we have the common KBase pattern of doing as little as possible in the IMPL file itself.

lib/execution_engine2/utils/clients.py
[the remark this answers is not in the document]
MrCreosote: Removed getters, no longer needed

lib/execution_engine2/execution_engine2Impl.py
MrCreosote: Oh I see. I'll make an APIHelper class that holds a reference to the config so we don't have to pass it around all over place and put the method there.

lib/execution_engine2/utils/clients.py
MrCreosote: Did that, this is now moot, no more getters

test/tests_for_auth/ee2_admin_mode_test.py
MrCreosote: done

test/tests_for_sdkmr/ee2_SDKMethodRunner_EE2Logs_test.py
MrCreosote: I don't think I understand the pattern. Is it 1 arg = no keyword arguments, > 1 arg = all keyword arguments, regardless of the function signature?

lib/execution_engine2/execution_engine2Impl.py
MrCreosote: done

test/tests_for_sdkmr/ee2_SDKMethodRunner_EE2Logs_test.py
bio-boris: Yea that's probably bad come to think of it.  The issue is calling functions with a large function signature with a bunch of positional arguments makes it hard to remember what is going on. Maybe we need to create an issue to address how we deal with that stylistically for the ee2 repo and save it to the contributing.md file

MrCreosote: Feel free to correct / elaborate: https://github.com/kbase/execution_engine2/issues/303

bio-boris: perfect!

lib/execution_engine2/utils/clients.py
bio-boris: Add dict['key'] vs dict.get('') to the style guide

bio-boris (approved):
Created new issues, questions answered, LGTM

MrCreosote:
Travis CI should be deleted per Boris

PR #306, opened 2021-02-05
Move boolean parsing to a separate module

MrCreosote:

Move the boolean parsing function into a separate module. Ran into import cycles when making other changes. Probably makes more sense to have it in a separate module anyway.

# Testing Instructions
- [X] Tests pass in travis and locally 

# Dev Checklist:

- [X] I have made corresponding changes to the documentation
- [?] My changes generate no new warnings - too many warnings to realistically check
- [X] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the travis build passes)

### DIFF
--- a/lib/execution_engine2/sdk/EE2StatusRange.py
+++ b/lib/execution_engine2/sdk/EE2StatusRange.py
@@ -6,6 +6,7 @@ from typing import Dict
 
 from bson import ObjectId
 
+from execution_engine2.utils.arg_processing import parse_bool
 from lib.execution_engine2.db.models.models import Job
 from lib.execution_engine2.exceptions import AuthError
 
@@ -187,7 +188,7 @@ class JobStatusRange:
         if ascending is None:
             return "+"
         else:
-            if self.sdkmr.parse_bool_from_string(ascending):
+            if parse_bool(ascending):
                 return "+"
             else:
                 return "-"
--- a/lib/execution_engine2/utils/arg_processing.py
+++ b/lib/execution_engine2/utils/arg_processing.py
@@ -13,19 +13,23 @@ def parse_bool(putative_bool: Union[str, bool, int, None]) -> bool:
 
     Raises ValueError if the value cannot be parsed.
     """
-    if putative_bool is None:
+    pb = putative_bool
+    if pb is None:
         return False
 
-    if isinstance(putative_bool, bool):
-        return putative_bool
+    if isinstance(pb, bool) or isinstance(pb, int) or isinstance(pb, float):
+        return bool(pb)
 
-    if isinstance(putative_bool, int):
-        return bool(putative_bool)
-
-    if isinstance(putative_bool, str):
-        if putative_bool.lower() == "true":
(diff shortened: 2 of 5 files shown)


MrCreosote:
The codacy and codeclimate complaints IMO are not useful or are incorrect

bio-boris (approved):
lgtm

lib/execution_engine2/utils/arg_processing.py
```diff
@@ -0,0 +1,31 @@
...
+    if putative_bool is None:
+        return False
+
+    if isinstance(putative_bool, bool):
+        return putative_bool
+
+    if isinstance(putative_bool, int):
+        return bool(putative_bool)
+
+    if isinstance(putative_bool, str):
```
bio-boris, line 25: Do we care about cases when it's a string not equal to "true" or "false" ? like "tru"?

bio-boris:
LGTM, but we might want to avoid our own function and use json.loads instead for t/f values?

lib/execution_engine2/utils/arg_processing.py
bio-boris: Ah never mind it throws the exception.. What about "1.0", "1", which is likely to come from the json rpc endpoints?

bio-boris (approved):

lib/execution_engine2/sdk/EE2StatusRange.py
```diff
@@ -187,7 +188,7 @@ def get_sort_order(self, ascending):
         if ascending is None:
             return "+"
         else:
-            if self.sdkmr.parse_bool_from_string(ascending):
+            if parse_bool(ascending):
```
bio-boris, line 191: we should see what the job browser is sending and see if that still works here

lib/execution_engine2/utils/arg_processing.py
MrCreosote: Added handling for stringified ints and floats

lib/execution_engine2/sdk/EE2StatusRange.py
MrCreosote: I can do this if you want, but it seems like the updated function should be able to handle absolutely anything thrown at it at this point. I'm assuming for now it's not necessary, but let me know if you'd still like me to check.

MrCreosote:
> we might want to avoid our own function and use json.loads/json.dumps instead for t/f values?

this is the json processing part of the function:

```
        if isinstance(json.loads(str_or_bool.lower()), bool):
            return json.loads(str_or_bool.lower())
```

For `isinstance(...)` to be true, the value of `str_or_bool.lower()` must be `"true" or "false"`. That means, AFAICT, this statement is equivalent:

```
if str_or_bool.lower() == "true"
    return True
if str_or_bool.lower() == "false"
    return False
```

and avoids loading a relatively heavyweight library, initializing any json processing engine state, etc.

We could eliminate a couple lines of code by making a static map `{"true": True, "false": False}` and checking and returning the map value *shrug*

Issue #287, opened 2021-01-21
Remove mongoengine context managers


MrCreosote:
They're unnecessary. For example:

```
        with self.sdkmr.get_mongo_util().mongo_engine_connection():
            self.logger.debug(job.to_mongo().to_dict())
            job.save()
```

PR #317, opened 2021-02-20
Fix workspace client access bug & unit test

MrCreosote:

Fixes a bug where a SDKMR.get_workspace() method was expected but was removed in a
previous PR.

Creates a unit test for the run() method where the bug occurred.

# Testing Instructions
- [X] Tests pass in Github Actions and locally 

# Dev Checklist:

- [?] My changes generate no new warnings - over 3k warnings currently
- [X] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the Github Actions build passes)

### DIFF
--- a/lib/execution_engine2/db/models/models.py
+++ b/lib/execution_engine2/db/models/models.py
@@ -111,6 +111,9 @@ class Meta(EmbeddedDocument):
     cell_id = StringField()
     status = StringField()
 
+    def __repr__(self):
+        return self.to_json()
+
 
 class CondorResourceUsage(EmbeddedDocument):
     """
@@ -147,6 +150,9 @@ class JobRequirements(EmbeddedDocument):
     disk = IntField()
     estimate = EmbeddedDocumentField(Estimate)
 
+    def __repr__(self):
+        return self.to_json()
+
 
 class JobInput(EmbeddedDocument):
     """
@@ -163,6 +169,9 @@ class JobInput(EmbeddedDocument):
     parent_job_id = StringField()
     requirements = EmbeddedDocumentField(JobRequirements)
     narrative_cell_info = EmbeddedDocumentField(Meta, required=True)
+
+    def __repr__(self):
+        return self.to_json()
 
 
 class JobOutput(EmbeddedDocument):
--- a/lib/execution_engine2/sdk/EE2Runjob.py
+++ b/lib/execution_engine2/sdk/EE2Runjob.py
@@ -20,7 +20,7 @@ from lib.execution_engine2.db.models.models import (
 )
 from lib.execution_engine2.sdk.EE2Constants import ConciergeParams
 from lib.execution_engine2.utils.CondorTuples import CondorResources
-from lib.execution_engine2.utils.KafkaUtils import KafkaCreateJob, KafkaQueueChange
+from execution_engine2.utils.KafkaUtils import KafkaCreateJob, KafkaQueueChange
 
 
 class JobPermissions(Enum):
@@ -44,7 +44,7 @@ class EE2RunJob:
     def __init__(self, sdkmr):
         self.sdkmr = sdkmr  # type: SDKMethodRunner
         self.override_clientgroup = os.environ.get("OVERRIDE_CLIENT_GROUP", None)
-        self.logger = self.sdkmr.logger
+        self.logger = self.sdkmr.get_logger()
 
     def _init_job_rec(
         self,
@@ -100,15 +100,14 @@ class EE2RunJob:
         job.job_input = inputs
         self.logger.debug(job.job_input.to_mongo().to_dict())
 
-        with self.sdkmr.get_mongo_util().mongo_engine_connection():
-            self.logger.debug(job.to_mongo().to_dict())
-            job.save()
+        self.logger.debug(job.to_mongo().to_dict())
+        job_id = self.sdkmr.save_job(job)
 
-        self.sdkmr.kafka_client.send_kafka_message(
-            message=KafkaCreateJob(job_id=str(job.id), user=user_id)
+        self.sdkmr.get_kafka_client().send_kafka_message(
+            message=KafkaCreateJob(job_id=job_id, user=user_id)
         )
 
-        return str(job.id)
+        return job_id
 
     def _get_module_git_commit(self, method, service_ver=None) -> Optional[str]:
         module_name = method.split(".")[0]
@@ -118,7 +117,7 @@ class EE2RunJob:
 
         self.logger.debug(f"Getting commit for {module_name} {service_ver}")
 
-        module_version = self.sdkmr.catalog_utils.catalog.get_module_version(
+        module_version = self.sdkmr.get_catalog_utils().get_catalog().get_module_version(
             {"module_name": module_name, "version": service_ver}
         )
 
@@ -194,7 +193,7 @@ class EE2RunJob:
         self._check_ws_objects(source_objects=params.get("source_ws_objects"))
         method = params.get("method")
         # Normalize multiple formats into one format (csv vs json)
-        normalized_resources = self.sdkmr.catalog_utils.get_normalized_resources(method)
+        normalized_resources = self.sdkmr.get_catalog_utils().get_normalized_resources(method)
         # These are for saving into job inputs. Maybe its best to pass this into condor as well?
         extracted_resources = self.sdkmr.get_condor().extract_resources(
             cgrr=normalized_resources
@@ -202,16 +201,16 @@ class EE2RunJob:
         # insert initial job document into db
 
         job_id = self._init_job_rec(
-            self.sdkmr.user_id, params, extracted_resources, concierge_params
+            self.sdkmr.get_user_id(), params, extracted_resources, concierge_params
         )
 
         params["job_id"] = job_id
-        params["user_id"] = self.sdkmr.user_id
-        params["token"] = self.sdkmr.token
+        params["user_id"] = self.sdkmr.get_user_id()
+        params["token"] = self.sdkmr.get_token()
         params["cg_resources_requirements"] = normalized_resources
 
         self.logger.debug(
-            f"User {self.sdkmr.user_id} attempting to run job {method} {params}"
+            f"User {self.sdkmr.get_user_id()} attempting to run job {method} {params}"
         )
 
         return PreparedJobParams(params=params, job_id=job_id)
@@ -249,8 +248,8 @@ class EE2RunJob:
         )
 
         self.update_job_to_queued(job_id=job_id, scheduler_id=condor_job_id)
-        self.sdkmr.slack_client.run_job_message(
-            job_id=job_id, scheduler_id=condor_job_id, username=self.sdkmr.user_id
+        self.sdkmr.get_slack_client().run_job_message(
+            job_id=job_id, scheduler_id=condor_job_id, username=self.sdkmr.get_user_id()
         )
 
         return job_id
@@ -374,23 +373,22 @@ class EE2RunJob:
         # TODO RETRY FOR RACE CONDITION OF RUN/CANCEL
         # TODO PASS QUEUE TIME IN FROM SCHEDULER ITSELF?
         # TODO PASS IN SCHEDULER TYPE?
-        with self.sdkmr.get_mongo_util().mongo_engine_connection():
-            j = self.sdkmr.get_mongo_util().get_job(job_id=job_id)
-            previous_status = j.status
-            j.status = Status.queued.value
-            j.queued = time.time()
-            j.scheduler_id = scheduler_id
-            j.scheduler_type = "condor"
-            j.save()
+        j = self.sdkmr.get_mongo_util().get_job(job_id=job_id)
+        previous_status = j.status
+        j.status = Status.queued.value
+        j.queued = time.time()
+        j.scheduler_id = scheduler_id
+        j.scheduler_type = "condor"
+        self.sdkmr.save_job(j)
 
-            self.sdkmr.kafka_client.send_kafka_message(
-                message=KafkaQueueChange(
-                    job_id=str(j.id),
-                    new_status=j.status,
-                    previous_status=previous_status,
-                    scheduler_id=scheduler_id,
-                )
+        self.sdkmr.get_kafka_client().send_kafka_message(
+            message=KafkaQueueChange(
+                job_id=str(j.id),
+                new_status=j.status,
+                previous_status=previous_status,
+                scheduler_id=scheduler_id,
             )
+        )
 
     def get_job_params(self, job_id, as_admin=False):
         """
--- a/lib/execution_engine2/sdk/EE2Runjob.py
+++ b/lib/execution_engine2/sdk/EE2Runjob.py
@@ -9,7 +9,7 @@ import time
 from enum import Enum
 from typing import Optional, Dict, NamedTuple, Union, List
 
-from lib.execution_engine2.db.models.models import (
+from execution_engine2.db.models.models import (
     Job,
     JobInput,
     Meta,
--- a/lib/execution_engine2/sdk/EE2Runjob.py
+++ b/lib/execution_engine2/sdk/EE2Runjob.py
@@ -117,8 +117,10 @@ class EE2RunJob:
 
         self.logger.debug(f"Getting commit for {module_name} {service_ver}")
 
-        module_version = self.sdkmr.get_catalog_utils().get_catalog().get_module_version(
-            {"module_name": module_name, "version": service_ver}
+        module_version = (
+            self.sdkmr.get_catalog_utils()
+            .get_catalog()
+            .get_module_version({"module_name": module_name, "version": service_ver})
         )
 
         git_commit_hash = module_version.get("git_commit_hash")
@@ -193,7 +195,9 @@ class EE2RunJob:
         self._check_ws_objects(source_objects=params.get("source_ws_objects"))
         method = params.get("method")
         # Normalize multiple formats into one format (csv vs json)
-        normalized_resources = self.sdkmr.get_catalog_utils().get_normalized_resources(method)
+        normalized_resources = self.sdkmr.get_catalog_utils().get_normalized_resources(
+            method
+        )
         # These are for saving into job inputs. Maybe its best to pass this into condor as well?
         extracted_resources = self.sdkmr.get_condor().extract_resources(
             cgrr=normalized_resources
--- a/lib/execution_engine2/sdk/SDKMethodRunner.py
+++ b/lib/execution_engine2/sdk/SDKMethodRunner.py
@@ -148,7 +148,7 @@ class SDKMethodRunner:
     # that instance variables cannot be detected by create_autospec with spec_set=True, and thus
     # cannot be mocked in a rigorous way. The danger of not using spec_set=True is that if a
     # mocked class's API changes, the unit tests will still pass. Thus the choice is between
-    # unpythonic getters or false positives in unit tests, and we choose the latter.
+    # unpythonic getters or false positives in unit tests, and we choose the former.
     # For more details: https://www.seanh.cc/2017/03/17/the-problem-with-mocks/
 
     def get_workspace(self) -> Workspace:
--- a/lib/execution_engine2/sdk/SDKMethodRunner.py
+++ b/lib/execution_engine2/sdk/SDKMethodRunner.py
@@ -12,6 +12,7 @@ import os
 import time
 from datetime import datetime
 from enum import Enum
+from logging import Logger
 
 import dateutil
 
@@ -34,6 +35,7 @@ from lib.execution_engine2.utils.KafkaUtils import KafkaClient
 from lib.execution_engine2.utils.SlackUtils import SlackClient
 from execution_engine2.utils.clients import UserClientSet
 from execution_engine2.utils.arg_processing import parse_bool
+from installed_clients.WorkspaceClient import Workspace
 
 
 class JobPermissions(Enum):
@@ -140,6 +142,62 @@ class SDKMethodRunner:
             self.condor = Condor(self.deployment_config_fp)
         return self.condor
 
+    # A note on getters:
+    # Getters are commonly described as unpythonic. However, accessing instance variables
+    # directly, rather than via getters, causes significant problems when mocking a class in
+    # that instance variables cannot be detected by create_autospec with spec_set=True, and thus
+    # cannot be mocked in a rigorous way. The danger of not using spec_set=True is that if a
+    # mocked class's API changes, the unit tests will still pass. Thus the choice is between
+    # unpythonic getters or false positives in unit tests, and we choose the latter.
+    # For more details: https://www.seanh.cc/2017/03/17/the-problem-with-mocks/
+
+    def get_workspace(self) -> Workspace:
+        """
+        Get the workspace client for this instance of SDKMR.
+        """
+        return self.workspace
+
+    def get_logger(self) -> Logger:
+        """
+        Get the logger for this instance of SDKMR.
+        """
+        # There's not really any way to meaningfully test this method without passing in the
+        # logger, which seems... overkill?
+        return self.logger
+
+    def get_catalog_utils(self) -> CatalogUtils:
+        """
+        Get the catalog utilities for this instance of SDKMR.
+        """
+        # TODO Unit test this method once catalog_utils can be mocked.
+        return self.catalog_utils
+
+    def get_kafka_client(self) -> KafkaClient:
+        """
+        Get the Kafka client for this instance of SDKMR.
+        """
+        # TODO Unit test this method once kafka_client can be mocked.
+        return self.kafka_client
+
+    def get_slack_client(self) -> SlackClient:
+        """
+        Get the Kafka client for this instance of SDKMR.
+        """
+        # TODO Unit test this method once slack_client can be mocked.
+        return self.slack_client
+
+    def get_user_id(self) -> str:
+        """
+        Get the user id of the user for this instance of SDKMR.
+        """
+        return self.user_id
+
+    def get_token(self) -> str:
+        """
+        Get the token of the user for this instance of SDKMR.
+        """
+        return self.token
+
     # Permissions Decorators    #TODO Verify these actually work     #TODO add as_admin to these
 
     def allow_job_read(func):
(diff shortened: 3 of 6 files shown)


lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -100,15 +100,14 @@ def _init_job_rec(
         job.job_input = inputs
         self.logger.debug(job.job_input.to_mongo().to_dict())
 
-        with self.sdkmr.get_mongo_util().mongo_engine_connection():
```
MrCreosote, line 103: See https://github.com/kbase/execution_engine2/issues/287

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -20,7 +20,7 @@
 )
 from lib.execution_engine2.sdk.EE2Constants import ConciergeParams
 from lib.execution_engine2.utils.CondorTuples import CondorResources
-from lib.execution_engine2.utils.KafkaUtils import KafkaCreateJob, KafkaQueueChange
+from execution_engine2.utils.KafkaUtils import KafkaCreateJob, KafkaQueueChange
```
ialarmedalien, line 23: ?? why no `lib.`?

lib/execution_engine2/sdk/SDKMethodRunner.py
```diff
@@ -140,6 +142,62 @@ def get_condor(self) -> Condor:
             self.condor = Condor(self.deployment_config_fp)
         return self.condor
 
+    # A note on getters:
+    # Getters are commonly described as unpythonic. However, accessing instance variables
+    # directly, rather than via getters, causes significant problems when mocking a class in
+    # that instance variables cannot be detected by create_autospec with spec_set=True, and thus
+    # cannot be mocked in a rigorous way. The danger of not using spec_set=True is that if a
+    # mocked class's API changes, the unit tests will still pass. Thus the choice is between
+    # unpythonic getters or false positives in unit tests, and we choose the latter.
```
ialarmedalien, line 151: 👍

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    # already a very large test. This may be something to be added later when needed.
+    sdkmr = create_autospec(SDKMethodRunner, spec_set=True, instance=True)
+    logger = create_autospec(Logger, spec_set=True, instance=True)
+    ws = create_autospec(Workspace, spec_set=True, instance=True)
+    catutils = create_autospec(CatalogUtils, spec_set=True, instance=True)
+    catalog = create_autospec(Catalog, spec_set=True, instance=True)
+    condor = create_autospec(Condor, spec_set=True, instance=True)
+    kafka = create_autospec(KafkaClient, spec_set=True, instance=True)
+    mongo = create_autospec(MongoUtil, spec_set=True, instance=True)
+    slack = create_autospec(SlackClient, spec_set=True, instance=True)
```
ialarmedalien, line 52: mockin' 'ell!

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    catutils.get_catalog.return_value = catalog
+    sdkmr.get_condor.return_value = condor
+    sdkmr.get_kafka_client.return_value = kafka
+    sdkmr.get_slack_client.return_value = slack
+    sdkmr.get_mongo_util.return_value = mongo
+    sdkmr.get_user_id.return_value = "someuser"
+    sdkmr.get_token.return_value = "tokentokentoken"
+    sdkmr.check_as_admin.return_value = True
+    sdkmr.save_job.return_value = "603051cfaf2e3401b0500982"
+    ws.get_object_info3.return_value = {"paths": [["1/2/3"], ["4/5/6"]]}
```
ialarmedalien, line 65: can you alphabetise these so it's easy for people to find whatever method they're looking for?

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    condor.run_job.return_value = SubmissionInfo("cluster42", {}, None)
+    retjob = Job()
+    retjob.id = ObjectId("603051cfaf2e3401b0500982")
+    retjob.status = "created"
+    mongo.get_job.return_value = retjob
+
+    # set up the class to be tested and run the method
+    rj = EE2RunJob(sdkmr)
+    params = {
+        "method": "lolcats.lol_unto_death",
```
ialarmedalien, line 85: I hope this is a real kbase app. If not, I propose it be put top of the wishlist for Q2 projects.

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    catutils.get_normalized_resources.assert_called_once_with("lolcats.lol_unto_death")
+    condor.extract_resources.assert_called_once_with(catalog_resources)
+    catalog.get_module_version.assert_called_once_with(
+        {"module_name": "lolcats", "version": "release"}
+    )
+
+    # initial job data save
+    expected_job = Job()
+    expected_job.user = "someuser"
+    expected_job.status = "created"
```
ialarmedalien, line 104: is it not possible to do something like 
```py
expected_job = Job( "user": "some user", "status": "created")
```
or do all these objects have to be constructed very laboriously?

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    jr.disk = "26"
+    ji.requirements = jr
+    ji.narrative_cell_info = Meta()
+    expected_job.job_input = ji
+    assert len(sdkmr.save_job.call_args_list) == 2
+    got_job = sdkmr.save_job.call_args_list[0][0][0]
+    assert_jobs_equal(got_job, expected_job)
+
+    kafka.send_kafka_message.assert_any_call(
+        KafkaCreateJob("someuser", "603051cfaf2e3401b0500982")
```
ialarmedalien, line 123: can you make `"603051cfaf2e3401b0500982"` into a variable since it's used numerous times?

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+
+def assert_jobs_equal(got_job: Job, expected_job: Job):
+    """
+    Checks that the two jobs are equivalent, except that the 'updated' fields are checked that
+    they're within 1 second of each other.
+    """
+    # I could not get assert got_job == expected_job to ever work, or even
+    # JobRequirements1 == JobRequirements2 to work when their fields were identical, so we
+    # do this disgusting hack instead. Note it will need to be updated any time a job field is
+    # added.
```
ialarmedalien, line 178: 😖

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    if not hasattr(got_job, "id"):
+        assert not hasattr(expected_job, "id")
+    else:
+        assert got_job.id == expected_job.id
+    assert got_job.user == expected_job.user
+    assert got_job.authstrat == expected_job.authstrat
+    assert got_job.wsid == expected_job.wsid
+    assert got_job.status == expected_job.status
```
ialarmedalien, line 186: all these repetitive statements are giving me refactoritis! Suggestions for refactoring if you can't do a straight object equality:
- make an explicit list of all the properties and do
```py
for property in property_list:
  assert got_job[property] == expected_job[property]
```
- list the keys of expected_job and do the same thing

It should be easy to omit `updated` from the check, whatever you choose to do.

test/tests_for_sdkmr/EE2Runjob_test.py
ialarmedalien: Actually, I'd do that for all the non-trivial values that you have use in the Job / JobInput / JobRequirements objects and the condor params object below.

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: I'm not sure why `lib` is in the path in the first place - the SDK sets the python path to `[repo]/lib` so it shouldn't be necessary. I'm a bit surprised it works at all. I imported without `lib` in my tests, and apparently when you have the same named tuple with different import lines either the mocks or the `Kafka*.__eq__` method gets confused and says tuples with identical contents aren't equal.

I'm gradually going around and removing the `lib`s when I need to.

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+        assert (
+            got_job.job_input.requirements.disk
+            == expected_job.job_input.requirements.disk
+        )
+        assert got_job.job_input.requirements.estimate == (
+            expected_job.job_input.requirements.estimate
+        )
+
+        # this fails, which should be impossible given all the fields above pass
+        # assert got_job.job_input.requirements == expected_job.job_input.requirements
```
ialarmedalien, line 243: are there other fields around that you're not accounting for?

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: I'm not super familiar with MongoEngine, but that's how it's done in their [tutorial](https://docs.mongoengine.org/tutorial.html#adding-data-to-our-tumblelog) as well as in the EE2 code, so my assumption is not.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: I don't think so: https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L151-L165

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Hmm, I'd have to use `getattr` or go into the instance dict for that, which I'm a bit meh about as it's a bit magic-y. How painful is the refactoritis?

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+    assert got_job.scheduler_type == expected_job.scheduler_type
+    assert got_job.scheduler_id == expected_job.scheduler_id
+    assert got_job.scheduler_estimator_id == expected_job.scheduler_estimator_id
+
+    if not got_job.job_input:
+        assert expected_job.job_input is None
+    else:
+        assert got_job.job_input.wsid == expected_job.job_input.wsid
+        assert got_job.job_input.method == expected_job.job_input.method
+        assert (
```
ialarmedalien, line 210: Did black add these brackets in? If not, have you ensured that `assert` is doing what you think it is?
e.g. see the 'Caveat #2' section of https://dbader.org/blog/python-assert-tutorial

What a great idea it was to make a programming language hypersensitive to whitespace. 😖

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: They're actually in the same order that the calls occur in the code, and I'd like to keep them that way unless there's a good reason not to. I'll add comments to that effect.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: done

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: > Did black add these brackets in? 

Yes. >:

test/tests_for_sdkmr/EE2Runjob_test.py
ialarmedalien: what's wrong with `getattr` or the instance dict? They are pretty standard methods.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: None of them could be interpreted as tuples so I think it's ok in this case, but that's a good warning. I'll add a comment to that effect.

MrCreosote: I deliberately made one fail just to be safe

MrCreosote: > far more sensible JavaScript.

I'm rolling

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: I'm not so sure they are. Directly accessing the instance dict is very rare in my experience, and IMO A Very Bad Thing In General. 

`getattr` is more common, but still pretty rare in my experience. But fine, I'll to it with `getattr`. Yeccch.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Huh, that's weird, there was only one comment when I wrote `done`. It's not done yet.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: MongoEngine does some weird stuff with the `__dict__`:

```
    print(repr(got_job))
    print(vars(got_job))
    print(got_job.__dict__)
    print(getattr(got_job, 'updated'))
```
results in
```
{"user": "someuser", "authstrat": "kbaseworkspace", "status": "created", "updated": 1614038718.449384, "job_input": {"method": "lolcats.lol_unto_death", "service_ver": "git5678", "source_ws_objects": ["1/2/3", "4/5/6"], "parent_job_id": "None", "requirements": {"clientgroup": "grotesquememlong", "cpu": 4, "memory": 32, "disk": 2600}, "narrative_cell_info": {}}, "child_jobs": [], "batch_job": false}
{'_cls': 'Job'}
{'_cls': 'Job'}
1614038718.449384
{'_cls': 'Job'}
```

MrCreosote: done

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: done

lib/execution_engine2/db/models/models.py
```diff
@@ -111,6 +111,9 @@ class Meta(EmbeddedDocument):
     cell_id = StringField()
     status = StringField()
 
+    def __repr__(self):
```
briehl, line 114: Good call on adding `__repr__`s.

lib/execution_engine2/sdk/SDKMethodRunner.py
```diff
@@ -140,6 +142,62 @@ def get_condor(self) -> Condor:
...
+        Get the workspace client for this instance of SDKMR.
+        """
+        return self.workspace
+
+    def get_logger(self) -> Logger:
+        """
+        Get the logger for this instance of SDKMR.
+        """
+        # There's not really any way to meaningfully test this method without passing in the
+        # logger, which seems... overkill?
```
briehl, line 165: :+1:

test/tests_for_sdkmr/EE2Runjob_test.py
briehl: Which test fails when calling this?

briehl (changes requested):
I think it's mostly fine. I'm concerned about why those tests are failing when they shouldn't be, though. Can you rewrite the equality checker to be more verbose and say something like "got XXX, expected YYY"?

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+        "scheduler_type",
+        "scheduler_id",
+        "scheduler_estimator_id",
+        "job_output",
+        "condor_job_ads",
+        "child_jobs",
+        "batch_job",
+    ]
+
+    _super_hacky_equals(got_job, expected_job, job_fields)
```
ialarmedalien, line 229: Can you rename this something more sensible? Initially I laughed at the name, but looking at the function, it's not actually hacky: you are just checking that a list of object attributes are equal.

test/tests_for_sdkmr/EE2Runjob_test.py
ialarmedalien: Did you try any of the `unittest` equality functions (e.g. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual) in case they give better diagnostics of the problem?

MrCreosote: > Which test fails when calling this?
The commented out assert fails

> Have you tried any of the unittest equality functions (e.g. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual) in case they give better diagnostics of the problem?

Since I'm not subclassing `TestCase`, which contains the `assertEquals()` method, I don't think I can, although if you know how to handle that without having to restructure the tests into a class I'll give it a go. In any case, in my experience, `pytest` with `assert`s give much better diagnostics than `unittest`.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Here I disagree - this function is super hacky. If a field is changed in the `Job` class this function will stop working correctly, possibly silently. I definitely want future programmers to know this is a function to be avoided if possible.

MrCreosote:
> Can you rewrite the equality checker to be more verbose and say something like "got XXX, expected YYY"

The equality checker in question is in the MongoEngine model super class, so not without an absolutely huge amount of trouble.

To be clear, the way I'm checking equality now (e.g. `getattr` and a list of fields to check) works, but is fragile and hacky. Trying to to something like
```
assert got_job == expected_job
```
Fails after setting any fields, and I was unable to figure out why after a substantial time investment. At this point throwing more time at it would just be reenacting the sunk cost fallacy

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -0,0 +1,268 @@
...
+            cell_info_fields,
+        )
+        # again, this fails, even though the fields are all ==
+        # assert got_job.job_input.narrative_cell_info == (
+        #     expected_job.job_input.narrative_cell_info)
+
+
+def _super_hacky_equals(obj1, obj2, fields):
+    for field in fields:
+        assert getattr(obj1, field) == getattr(obj2, field), field
```
ialarmedalien, line 268: To aid future troubleshooting, I would either add in a comment to the effect of  "An `AttributeError` will be raised if the field is not present in the object" or add in a test that both objects have those attributes.

test/tests_for_sdkmr/EE2Runjob_test.py
ialarmedalien: The `assert_job_equals` function is hacky, but `_super_hacky_equals` is not. You've already put in a disclaimer about why `assert_job_equals` is necessary and why it's horrible. Calling a function that uses a `for` loop and the `getattr` function  "hacky" is only going to invite confusion, since both are core to the functionality of python3.

Again, the reason that the function is required is hacky, which you have explained. The function itself is not. Please rename it to avoid confusion.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Added the comment

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: >  Calling a function that uses a for loop and the getattr function "hacky" is only going to invite confusion, since both are core to the functionality of python3.

It's not hacky because of `getattr` or the `for` loop, it's hacky because it's implementing an equals function that is fragile and virtually guaranteed to break sometime in the future. No one should use that equals function unless there's no other alternative. The case that the name is appropriate is very strong here.

MrCreosote: Actually, I have an idea that may make us both happy. Give me a sec

test/tests_for_sdkmr/EE2Runjob_test.py
briehl: Sure, I see the commented out assert, but you call the `assert_jobs_equal` function twice - line 120 and line 155. Do they both always fail? Or just one or the other? It could have to do with how you're mocking. The `expected_job` has its `narrative_cell_info` set to a `Meta()` object, while `got_job` might have it as an `EmbeddedDocumentField` wrapped around a `Meta()`. If they don't have an `__eq__` function that does duck-typing against its values, but also takes in the class, it might get wiggy when you compare the outer classes like that. I don't know enough about MongoEngine to know what's happening.

Also one or the other class might have extra attributes set.

MrCreosote: > you call the assert_jobs_equal function twice - line 120 and line 155. Do they both always fail?

Neither fail, both pass.

MrCreosote: > The expected_job has its narrative_cell_info set to a Meta() object, while got_job might have it as an EmbeddedDocumentField wrapped around a Meta().

If that were true, comparing the fields of the Meta() object to the fields of the EDF wrapper should cause the test to fail.

MrCreosote: >  I don't know enough about MongoEngine to know what's happening

MrCreosote: > Also one or the other class might have extra attributes set.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: done

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: here's some more fun:

```
            print(vars(got_job.job_input.requirements))
            print(got_job.job_input.requirements.estimate)
            print(repr(got_job.job_input.requirements.estimate))
            print(type(got_job.job_input.requirements.estimate))
            print(vars(expected_job.job_input.requirements))
            print(expected_job.job_input.requirements.estimate)
            print(repr(expected_job.job_input.requirements.estimate))
            print(type(expected_job.job_input.requirements.estimate))
    
            # this fails, which should be impossible given all the fields above pass
>           assert got_job.job_input.requirements == expected_job.job_input.requirements
```

Results in:

```
{'_cls': 'JobRequirements'}
None
None
<class 'NoneType'>
{'_cls': 'JobRequirements'}
None
None
<class 'NoneType'>
```

So unless there's some invisible field set by the superclass I'm stumped.

MrCreosote: There are only the 5 fields in JobRequirements: https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L139-L148

MrCreosote: Neither have an `id` field

MrCreosote: I got the embedded documents to all pass normally - it was a `lib` in an import line again.

I still can't get the outermost `Job` class to pass an equality check even when the fields are identical.

MrCreosote: Ok, so far I've
- confirmed equals for the the `job_input` field works - this is the only field that is or contains embedded objects
- set the job_input field in `got_job` and `expected_job` to `None`
- confirmed the tests still fail when trying to use a direct assert even when the fields appear identical:
```
>       assert got_job == expected_job
E       assert {"user": "someuser", "authstrat": "kbaseworkspace", "status": "created", "updated": 1614120410.3491368, "child_jobs": [], "batch_job": false} == {"user": "someuser", "authstrat": "kbaseworkspace", "status": "created", "updated": 1614120410.3491368, "child_jobs": [], "batch_job": false}
E         +{"user": "someuser", "authstrat": "kbaseworkspace", "status": "created", "updated": 1614120410.3491368, "child_jobs": [], "batch_job": false}
E         -{"user": "someuser", "authstrat": "kbaseworkspace", "status": "created", "updated": 1614120410.3491368, "child_jobs": [], "batch_job": false}
```
- confirmed the types and import paths for the Job objects are the same
- confirmed the types and values of all the fields in the field list are the same

At this point I'm out of ideas.

MrCreosote: Well, here's the problem: 
https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/base/document.py#L284-L297

MrCreosote: The hierarchy is `Job` -> `Document` -> `BaseDocument`. Neither `Job` nor `Document` have an `__eq__` method. In contrast, `JobInput` and the other embedded classes inherit from `EmbeddedDocument`: https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/document.py#L85-L88

MrCreosote: Oh and all these classes use `__slots__` hence no dict

ialarmedalien (approved):
👍

briehl (approved):

MrCreosote:
It kills me to know that all this work figuring out wtf was going on here is going to be squashed into nothingness

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: Yes, in that link you posted you can see that you can do something like
```
>>> class User(Document):
...     email = StringField(required=True)
...     first_name = StringField(max_length=50)
...     last_name = StringField(max_length=50)
...
>>> u = User(email='example@example.com', first_name='bob', last_name='dole')
>>> u
<User: User object>
```

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: It is possible to get a dict like so

job.to_mongo().to_dict() from both job objects and then compare the dictionaries

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Ha, I guess my eyes were drawn to the big old note and that confirmed what I suspected so I moved on

PR #325, opened 2021-03-05
Add argument / parameter checkers

MrCreosote:

Adds argument / parameter checking helper code for use in future PRs.

# Testing Instructions
- [X] Tests pass in Github Actions and locally 

# Dev Checklist:

- [?] My changes generate no new warnings - over 3k warnings
- [X] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the Github Actions build passes)

### DIFF
--- a/lib/execution_engine2/utils/arg_processing.py
+++ b/lib/execution_engine2/utils/arg_processing.py
@@ -8,7 +8,7 @@ import datetime
 import unicodedata
 from execution_engine2.exceptions import IncorrectParamsException
 
-T = TypeVar('T')
+T = TypeVar("T")
 
 
 def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
@@ -49,23 +49,22 @@ def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
 
 
 def not_falsy(item: T, item_name: str) -> T:
-    '''
+    """
     Check if a value is falsy and throw and exception if so.
     :param item: the item to check for falsiness.
     :param item_name: the name of the item to include in any exception.
     :raises ValueError: if the item is falsy.
     :returns: the item.
-    '''
+    """
     if not item:
-        raise ValueError(f'{item_name} cannot be a value that evaluates to false')
+        raise ValueError(f"{item_name} cannot be a value that evaluates to false")
     return item
 
 
 def not_falsy_in_iterable(
-        iterable: Optional[Iterable[T]],
-        name: str,
-        allow_none: bool = False) -> Optional[Iterable[T]]:
-    '''
+    iterable: Optional[Iterable[T]], name: str, allow_none: bool = False
+) -> Optional[Iterable[T]]:
+    """
     Check that an iterable is not None and contains no falsy items. Empty iterables are accepted.
     :param iterable: the iterable to check.
     :param name: the name of the iterable to be used in error messages.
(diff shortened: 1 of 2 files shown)


lib/execution_engine2/utils/arg_processing.py
```diff
@@ -33,4 +38,124 @@ def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
...
+    """
+    # make public if needed
+    # See https://stackoverflow.com/questions/4324790/removing-control-characters-from-a-string-in-python  # noqa: E501
+    for c in string:
+        if unicodedata.category(c)[0] == "C":
+            return True
+    return False
+
+
+def _no_control_characters(string: str, name: str) -> str:
```
ialarmedalien, line 101: Does this need to be in a separate method from `_contains_control_characters`? Are there situations where you are likely to use the latter but not the former?

lib/execution_engine2/utils/arg_processing.py
```diff
@@ -33,4 +38,124 @@ def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
...
             return False
 
-    raise ValueError(f"{pb} is not a boolean value")
+    raise IncorrectParamsException(f"{pb} is not a boolean value")
+
+
+# The remaining methods are ported from
+# https://github.com/kbase/sample_service/blob/master/lib/SampleService/core/arg_checkers.py
+# with slight changes.
+# Should probably make a package or see if there are equivalent 3rd party functions at some point.
```
ialarmedalien, line 47: were the tests also ported over?

MrCreosote: yes, with modifications

lib/execution_engine2/utils/arg_processing.py
```diff
@@ -33,4 +38,124 @@ def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
...
+    """
+    Check that a string meets a set of criteria:
+    - it is not None or whitespace only (unless the optional parameter is specified)
+    - it contains no control characters
+    - (optional) it is less than some specified maximum length
+    :param string: the string to test.
+    :param name: the name of the string to be used in error messages.
+    :param max_len: the maximum length of the string.
+    :param optional: True if no error should be thrown if the string is None.
+    :returns: the stripped string or None if the string was optional and None or whitespace only.
```
briehl, line 128: I see the string returns the stripped version, but should it? I would almost expect that to either be an option, or just to return the same string I sent to it.

lib/execution_engine2/utils/arg_processing.py
MrCreosote: Way back when I wrote these originally, I had them split up because I did use them separately. However, that situation went away and so I made them private. Hence the docs on internal methods, which I usually don't do.

I could imagine a situation where you might want to scan a set of strings for control characters and report to the user which ones are bad rather than throwing an exception on the first one, or something along those lines, so there's at least a theoretical use case where splitting the method makes sense.

The functions are easy enough to grok that I don't think having them split up adds significant cognitive burden, and since they're not part of the API they don't add testing burden, so my preference is to leave things as is. I don't feel particularly strongly about it though.

briehl (changes requested):
Just a couple of typos, and one question.

lib/execution_engine2/utils/arg_processing.py
MrCreosote: After a whooole lot of experience using these functions it absolutely should. I've used functions like this hundreds of times in code since 2016 or so and I always want the strip. The use case for most of these functions is checking user inputs for functions or constructors where whitespace around strings is to be ignored.

lib/execution_engine2/utils/arg_processing.py
```diff
@@ -33,4 +38,124 @@ def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
...
+# The remaining methods are ported from
+# https://github.com/kbase/sample_service/blob/master/lib/SampleService/core/arg_checkers.py
+# with slight changes.
+# Should probably make a package or see if there are equivalent 3rd party functions at some point.
+# Although if you want to use custom exceptions as here that won't work
+
+
+def not_falsy(item: T, item_name: str) -> T:
+    """
+    Check if a value is falsy and throw and exception if so.
```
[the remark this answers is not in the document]
MrCreosote: I have this fixed locally, it'll show up in a future PR if that's ok.

lib/execution_engine2/utils/arg_processing.py
```diff
@@ -33,4 +38,124 @@ def parse_bool(putative_bool: Union[str, bool, int, float, None]) -> bool:
...
+    - (optional) it is less than some specified maximum length
+    :param string: the string to test.
+    :param name: the name of the string to be used in error messages.
+    :param max_len: the maximum length of the string.
+    :param optional: True if no error should be thrown if the string is None.
+    :returns: the stripped string or None if the string was optional and None or whitespace only.
+    :raises IncorrectParamsException: if the string is None, whitespace only, too long, or
+        contains illegal characters.
+    """
+    # See the IDMapping service if character classes are needed.
```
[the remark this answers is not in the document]
MrCreosote: Again I have this fixed locally and you can expect it in a future PR unless this is important enough it needs to be fixed immediately.

briehl (approved):

ialarmedalien (approved):
LGTM

PR #367, opened 2021-04-27
DATAUP-433 Create 002-Retry_endpoint.md

bio-boris:

* Please include a summary of the change and which issue is fixed. 
* Please also include relevant motivation and context.
* List any dependencies that are required for this change.

- [ ] Added the Jira Ticket to the title of the PR e.g. (DATAUP-69 Adds a PR template)

# Testing Instructions
- [ ] Changes available by spinning up a local test suite and doing X

# Dev Checklist:

- [ ] My code follows the guidelines at https://sites.google.com/truss.works/kbasetruss/development
- [ ] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the Github Actions build passes)


docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+* `-` Loss of information about job runs, and ability to infer relationships between parents and child jobs. 
+* `-` Loss of control of jobs, such as the ability to restrict a job's running
+* `-` Wrong version of app will run if the app was updated after job completion, and a version tag rather than a git commit was provided
+* `-` Increase complexity of `run_job*` methods
+ 
+### Re-writing `run_job` to address the aforementioned deficiencies without refactoring
+* `+` Solves most requirements, but 
+* `-` Adds more complexity to `run_job` methods
+* `-` Increase difficulty in maintaining and testing `run_job` method
+* `-` Wrong version of app will run if the app was updated after job completion, and a version tag rather than a git commit was provided
```
ialarmedalien, line 63: what happens with this at the moment -- does the job record store the version tag? the git commit? both?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+Pending decision
+
+## Pros and Cons of the Alternatives
+
+### Not book-keeping, or doing minimal book-keeping and calling `run_job` multiple times
+* `+` Can re-use existing endpoints without any additional work
+* `+` Less api endpoints to manage
+* `-` Issues on re-rendering/regenerating a cell based on just the job record
+* `-` Loss of information about job runs, and ability to infer relationships between parents and child jobs. 
+* `-` Loss of control of jobs, such as the ability to restrict a job's running
```
ialarmedalien, line 55: Can you clarify this a bit -- give a more concrete example of what you mean by "ability to restrict a job's running"?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
+# Retry Endpoint
+
+Date: 2021-04-27
+
+The current requirement for the Batch/Bulk UI is to be able to retry jobs. Using a job id, it should be possible to get information from the database.
+
+The current implementation of retry is to run jobs using the `run_job` or `run_job_batch` endpoint. This is not adequate due to the following deficiencies:
+
+
+* Lack of book-keeping for the parent job and the child job relationship: Launching a new job will not update the relationship between those jobs
```
ialarmedalien, line 10: Make this more specific -- launching a job that should be a child of a given parent job does not update the parent job to add the new child ID (I assume that's the case, anyway)

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+* Lack of book-keeping for number of retries for a particular job / set of job inputs
+* Lack of ability to launch multiple jobs using the `run_job_batch` endpoint without creating a new parent job
+* Lack of ability to ensure that the proper catalog version /git commit of an app is used from the front end based on a tag, such as "beta/dev/release"
+* Lack of ability to specify which retries succeeded and which ones failed during submit time.
```
ialarmedalien, line 14: This needs more explanation -- what is "submit time"?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+Pending
+
+* Link to spec file with inputs and outputs for the retry endpoint goes here
+* Link to Jira Ticket with business logic documentation for success and cancel cases
+
+### Possible additional things to think about
+* Creating sets of objects and what to do at the end of a batch run
+* What to do about a set if a child task fails during processing 
+* Convenience endpoints that operate on the parent_job_id or list of child job ids may be out of scope (e.g. cancel all jobs with a certain status)
+
```
ialarmedalien, line 43: - should retries be limited to jobs with certain statuses? e.g. it would make little sense to retry a job that completed successfully or a job that is queued or running.
- should the number of retries of a job be limited, and if so, where? e.g. a `max_retries` field in the parent job? wait and see whether people attempt to rerun jobs that have already failed nine zillion times?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+## Consequences
+
+Pending decision
+
+## Pros and Cons of the Alternatives
+
+### Not book-keeping, or doing minimal book-keeping and calling `run_job` multiple times
+* `+` Can re-use existing endpoints without any additional work
+* `+` Less api endpoints to manage
+* `-` Issues on re-rendering/regenerating a cell based on just the job record
```
ialarmedalien, line 53: can you explain this a bit more?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+* `+` Same as above, but if you are refactoring, you might as well have a retry endpoint, and clean out/decouple `run_job` endpoint from having so many features and branching logic
+
+### Creating `retry` endpoint to address the aforementioned deficiencies with some refactoring
+* `+` Decrease coupling between `run_job` and retry functionality, possibly making testing and development easier
+* `+` Faster development than a full refactor
+* `-` Faster development than a full refactor, but creates technical debt, might have to update both `run_job` and `retry` each time a change is made
+* `-` Extra endpoint to manage
+
+
+### Creating `retry` endpoint to address the aforementioned deficiencies with full refactoring where run_job functions are split out into their own functions
```
ialarmedalien, line 77: 🎉

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,81 @@
...
+
+Pending decision
+
+## Pros and Cons of the Alternatives
+
+### Not book-keeping, or doing minimal book-keeping and calling `run_job` multiple times
+* `+` Can re-use existing endpoints without any additional work
+* `+` Less api endpoints to manage
+* `-` Issues on re-rendering/regenerating a cell based on just the job record
+* `-` Loss of information about job runs, and ability to infer relationships between parents and child jobs.
```
ialarmedalien, line 54: client would be responsible for maintaining records to infer relationship between parent and child jobs

docs/adrs/002-Retry_endpoint.md
ialarmedalien: - how to prevent incorrect parent-child relationships being created -- should the client be allowed to specify a parent ID? Is it currently possible to add a new child to a parent job if the child is a new job, rather than an existing job ID / set of params that is being rerun?

docs/adrs/002-Retry_endpoint.md
bio-boris: The job record stores the git commit.

docs/adrs/002-Retry_endpoint.md
bio-boris: Thanks, fixed it to say restrict based on Retries/Failures

docs/adrs/002-Retry_endpoint.md
bio-boris: Comment to be added into the document somehow: The job first needs to be submitted to ee2_runjob. It can fail there. Afterwards, it gets submitted to condor, it can fail there too. Currently those two happen at the same time, but they are supposed to happen in a thread at some point, so the current single point of submit time will become two separate points of submission.
Once the job begins running, it can fail at Run Time.

docs/adrs/002-Retry_endpoint.md
bio-boris: * If you want to make sure the analysis is reproducible, you may want to retry that job, but then you would lose it in the UI, so you probably wouldn't want to do that. If you suspect the job is forever queued or forever stuck running, you may want to retry it. 
* Maybe have a "retry" field in the child job, and once it has reached a threshold, post an alert that says "Are you really sure you want to retry this job, it has failed "X" times" ?
* Any child can claim a parent right now currently, as per kbparallels. It is not currently possible to add a new child job to a parent job in any way.

docs/adrs/002-Retry_endpoint.md
bio-boris: Not sure I understand. Currently, run_job doesn't keep track of child jobs for the parent, and doesn't keep a record of which jobs we want and which ones we don't want
The parent record has a list of child jobs, and it's not currently being updated upon adding or removing child jobs.

docs/adrs/002-Retry_endpoint.md
bio-boris: We can in theory attempt to re-create a cell solely by the job record.

docs/adrs/002-Retry_endpoint.md
bio-boris: Updated it, thanks

docs/adrs/002-Retry_endpoint.md
ialarmedalien: So this point about the version tag is not really relevant to the discussion.

docs/adrs/002-Retry_endpoint.md
ialarmedalien: The comment was just clarifying that whatever is using ee2 (the client) would have to maintain the list of job IDs.

docs/adrs/002-Retry_endpoint.md
ialarmedalien: if all the child job IDs have been saved, this would not be a problem.

docs/adrs/002-Retry_endpoint.md
ialarmedalien: > If you suspect the job is forever queued or forever stuck running, you may want to retry it.
Is it possible that a job could get stuck in the queued or running states? Are there periodic checks done by ee2 to see if a job is still active?

ialarmedalien: > Any child can claim a parent right now currently, as per kbparallels. It is not currently possible to add a new child job to a parent job in any way.

That seems contradictory -- if any child can claim a parent, you could just create a new job and give it whatever parent you want to add it to.

I would be in favour of preventing external modules from creating parent-child relationships where possible, to reduce the scope for crazy job structures like circular parent-child relationships or huge job trees.

bio-boris: * No there are no checks done by ee2 to see if a job is still active. There are cases when the job gets stuck where that check wouldn't work either.

The parent job has a list of child jobs 
```
Parent
child_jobs = [1,2,3]
```

You can claim a parent, but the child_jobs array doesn't get updated. Just because you claim the job is your parent, doesnt mean the parent recognizes you as your child.

It seems to be in scope to support apps like metabat, which does exactly that, so we can't prevent that easily right now.

docs/adrs/002-Retry_endpoint.md
bio-boris: RIght now run_job as it currently stands, doesn't save number of retries for a job, update the parent job,  and the narrative wouldn't know which jobs have been retried, and thus to show a row for, and which ones not to. Thus it would either not render the retried jobs because it doesn't know about them, or if all of them are append to the list of child jobs, it would render all of the retried jobs as well.   The purpose of this section is to explain some deficiencies  with using run_job as it is today.

docs/adrs/002-Retry_endpoint.md
bio-boris: Got it, I'll add it as a negative

docs/adrs/002-Retry_endpoint.md
bio-boris: Yes, but the job is currently submitted via the code cell/narrative and that doesn't always store git commit. It stores "Dev|Beta|release|git commit". Submitting the job again with just the info from the browser has a chance to choose the wrong commit.

briehl: Another possible issue here is that there's multiple ways to interpret "version" of an app.
If the app's in a module with the release tag, it has at least one semantic version. There should also be a git commit of what's deployed.
If it's either in beta or dev, there's no official release, so no semantic version. There's always a git commit, though.

It might be best to always submit a git commit for the module, maybe?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,87 @@
...
+* `-` Increase complexity of `run_job*` methods
+* `-` The client will have to keep track of the child_job relationship, so that info is lost once the client is terminated
+ 
+### Re-writing `run_job` to address the aforementioned deficiencies without refactoring
+* `+` Solves most requirements, but 
+* `-` Adds more complexity to `run_job` methods
+* `-` Increase difficulty in maintaining and testing `run_job` method
+* `-` Wrong version of app will run if the app was updated after job completion, and a version tag rather than a git commit was provided
+* `-` Inefficient job submission
+* `-` Possibly Insufficient error handling
```
briehl, line 71: What could be insufficient? Are there errors that wouldn't be able to be handled by the endpoint?

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,87 @@
...
+### Re-writing `run_job` to address the aforementioned deficiencies without refactoring
+* `+` Solves most requirements, but 
+* `-` Adds more complexity to `run_job` methods
+* `-` Increase difficulty in maintaining and testing `run_job` method
+* `-` Wrong version of app will run if the app was updated after job completion, and a version tag rather than a git commit was provided
+* `-` Inefficient job submission
+* `-` Possibly Insufficient error handling 
+
+### Re-writing `run_job/run_job_batch` to address the aforementioned deficiencies with some refactoring
+* `+` Same as above, but if you are refactoring, you might as well have a retry endpoint, and clean out/decouple `run_job` endpoint from having so many features and branching logic
```
briehl, line 74: 👍

docs/adrs/002-Retry_endpoint.md
bio-boris: > It might be best to always submit a git commit for the module, maybe?
+1

docs/adrs/002-Retry_endpoint.md
bio-boris: Darn, I can't remember exactly what I was thinking, maybe I should remove this... There are some error cases that are hard to capture when handling multiple jobs, and at different points of failure (submit time at ee2, submit time at condor, run time at condor) 
The return structure is not suited to handle differentiating between different types of error messages as well right now, but it could be possible to create a more complicated return structure.

docs/adrs/002-Retry_endpoint.md
ialarmedalien: should have known better than to think that the narrative would always act in the same way! 😖

docs/adrs/002-Retry_endpoint.md
ialarmedalien: Do you know if there are any other apps that do this? It might be easier to edit metabat and to ensure that ee2 is in charge of all job relationships than to have to work around a few exceptions.

docs/adrs/002-Retry_endpoint.md
ialarmedalien: This is really just the same point as is mentioned below, that whatever the client is (the narrative, some third-party app, whatever) could do a lot of record keeping about the relationships between the different jobs; if it didn't, there wouldn't be a way to reconstruct the job relationships just using ee2 data. I think I would say:

* `-` Narrative would be forced to maintain detailed job records to be able to re-render or regenerate a batch job cell where jobs within the batch had been retried.

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,87 @@
...
+* `+` Faster development than a full refactor
+* `-` Faster development than a full refactor, but creates technical debt, might have to update both `run_job` and `retry` each time a change is made
+* `-` Extra endpoint to manage
+
+
+### Creating `retry` endpoint to address the aforementioned deficiencies with full refactoring where run_job functions are split out into their own functions
+* `+` Decrease coupling between `run_job` and retry functionality, possibly making testing and development easier
+* `+` Increase DRYNESS of the code
+* `+` Allows retry to benefit from changes to `run_job` 
+* `-` Slower development for a full refactor, but decreases technical debt
```
MrCreosote, line 87: Before we make a decision on implementation details like refactoring vs no refactoring, I feel like we should figure out the behavior we want a job retry to have. AJ brought up some good points here: https://github.com/kbase/execution_engine2/pull/367/files#r621719800

Behavior questions, incorporating AJ's notes and my own (AJ let me know if I missed any):

* Should we track a retry count?
* Should users see this retry count?
* Are retried jobs saved in some sort of data structure linking them, possibly indirectly, to the parent job or are they orphaned?
* If the former, is the retry relationship linear or a tree? E.g. what happens if there are two simultaneous calls to retry a job?
* Should it be at least theoretically possible to see the list of retried jobs in order?
* Should there be a maximum retry count? Or a warning that more retries are not likely to help?
* Can a job in states other than failed or canceled be retried? Or should the user be required to cancel a job before it can be retried?

MrCreosote: Once we make those decisions, to me the next step would be figuring out the data model in the EE2 DB we'd want to use to represent the state of retried jobs and how to make changes to that model atomically, or if that's not possible, to make the model resilient to unexpected server / network / db outages. Next the operations on that model that occur during a retry. That'll tell us how much would have to change in the EE2RunJob class, assuming that handles retries as well.

MrCreosote: Does this make sense to people or am I off base?

MrCreosote (commented):
I'm going to keep reviewing but these are pretty large size comments, if that makes sense, so I wanted to post them and then move on.

docs/adrs/002-Retry_endpoint.md
MrCreosote: I'm with AJ on thinking it's scary that any process can claim any job as a parent job for a job that it submits.

For retries for us, I don't think that's necessary as the parent job can just be looked up from the job record created by `run_job_batch`.

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,87 @@
...
+* `-` Adds more complexity to `run_job` methods
+* `-` Increase difficulty in maintaining and testing `run_job` method
+* `-` Wrong version of app will run if the app was updated after job completion, and a version tag rather than a git commit was provided
+* `-` Inefficient job submission
+* `-` Possibly Insufficient error handling 
+
+### Re-writing `run_job/run_job_batch` to address the aforementioned deficiencies with some refactoring
+* `+` Same as above, but if you are refactoring, you might as well have a retry endpoint, and clean out/decouple `run_job` endpoint from having so many features and branching logic
+
+### Creating `retry` endpoint to address the aforementioned deficiencies with some refactoring
```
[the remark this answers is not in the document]
briehl: I think the goal here was just to record the different options we had and rejected, in case someone ever wants to come back to it.

docs/adrs/002-Retry_endpoint.md
bio-boris: I think these questions are better suited on the retry ticket. From my understanding, the point of this ADR was to justify a retry endpoint, and some of the reasons it better to have that then to use the existing run_job endpoint as is.

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,87 @@
...
+* `+` Decrease coupling between `run_job` and retry functionality, possibly making testing and development easier
+* `+` Faster development than a full refactor
+* `-` Faster development than a full refactor, but creates technical debt, might have to update both `run_job` and `retry` each time a change is made
+* `-` Extra endpoint to manage
+
+
+### Creating `retry` endpoint to address the aforementioned deficiencies with full refactoring where run_job functions are split out into their own functions
```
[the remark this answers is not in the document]
bio-boris: I think this alternative is  meant to illustrate the benefit of refactoring, versus not refactoring, and possibly address the coupling between running jobs and run_job, run_job_batch, and retry_job(s) , while mentioning that it takes more development time and potentially reduces debt.

docs/adrs/002-Retry_endpoint.md
bio-boris: Because it takes a long time to submit multiple jobs, and we want a list of job ids right away

docs/adrs/002-Retry_endpoint.md
MrCreosote: My understanding was that we had decided we *did* want a retry endpoint, and the purpose of the ADR was to figure out how it should work - that's what the notes from the engineering checkin say anyway.

docs/adrs/002-Retry_endpoint.md
MrCreosote: Ok, so you make the records in the DB and then start submitting to Condor, I see. That'll take a pretty big rewrite of `_run` though.

docs/adrs/002-Retry_endpoint.md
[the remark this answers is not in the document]
MrCreosote: >  it takes more development time

I think my point is that it might not actually take much more development time and therefore there's no real tradeoff here. 

Could be wrong of course.

docs/adrs/002-Retry_endpoint.md
sychan: I am with Gavin on this, we are agreed on the desirability of the retry endpoint, but want to document the decision and discuss the details of what the retry endpoint will do, and discuss the effort involved for different options.

As I mentioned, as the product owner, I find our ability to deliver functionality to be pretty awful. We have invested so much effort in refactoring that its killed our timeline - we started in late July, and it is now almost May with no functioning bulk uploader, which was just the first deliverable.

If we are going to refactor, we need to be able to do it in a timely fashion, and have it not kill the schedule any more than it has. I want to see the estimate for a quick and dirty solution that implements a proposed retry endpoint, that can be deployed ASAP, and then once the API contract has been established, and the functional MVP is done, we begin the cleanup of the backend code.

Note that this is *NOT* business as usual, the usual way we do this is the nasty MVP gets deployed and then we don't go back until much later. Here, we get the API working so that it doesn't block dependencies, and we immediately start the refactoring. The refactor needs to be broken down into smallish chunks of ~3 days estimated work, and each merge should maintain functionality and incrementally improve the codebase. Tasks that take more than a couple of days are more likely to be far off in their estimate and this is how we mitigate the risk of poor estimation.

ialarmedalien: > Should it be at least theoretically possible to see the list of retried jobs in order?

If this is not done directly, it could be done by ordering the jobs by creation time

ialarmedalien: > Should there be a maximum retry count? Or a warning that more retries are not likely to help?

This is something that could be enforced by the client if it isn't implemented by ee2.

ialarmedalien: > Are retried jobs saved in some sort of data structure linking them, possibly indirectly, to the parent job or are they orphaned?
> If the former, is the retry relationship linear or a tree? E.g. what happens if there are two simultaneous calls to retry a job?

If the retried jobs are not linked to the parent job, we lose a lot of the power of having a retry endpoint. To simplify things, I think it would be best for only one execution of a given set of parameters to be permitted at a time. This implies that ee2 is able to distinguish unique jobs in a batch; the simplest way to do this would be to generate a hash from each set of params [or, if there is a unique param, such as output object name, use that], and reject any duplicated param sets.

Another possible alternative would be tracking the position of the job in the initial input set and having a field that links the rerun to the original job.

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,95 @@
+# Retry Endpoint
```
bio-boris, line 1: TODO add info from https://github.com/kbase/execution_engine2/pull/383#discussion_r640836726

docs/adrs/002-Retry_endpoint.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+### Possible additional things to think about
+* Creating sets of objects and what to do at the end of a batch run
+* What to do about a set if a child task fails during processing 
+* Convenience endpoints that operate on the parent_job_id or list of child job ids may be out of scope (e.g. cancel all jobs with a certain status)
+
+## Consequences
+
+Pending decision
```
MrCreosote, line 59: I guess this needs to be updated since a decision has been made?

MrCreosote (changes requested):
This review is for ADR #1, which looks fine to me other than one part that needs to be filled in post decision.

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -0,0 +1,118 @@
...
+    Should it be at least theoretically possible to see the list of retried jobs in order? (It is possible by sorting on creation date)
+    Should there be a maximum retry count? Or a warning that more retries are not likely to help?  (Unknown TBD)
+    Can a job in states other than failed or canceled be retried? Or should the user be required to cancel a job before it can be retried? (Job must be in Error/Cancel state)
+
+
+# Work estimation
+Priority descending
+* Non blocking job submission for submitting multiple jobs, possibly via using `run_job_batch` (requires refactor of run_job_batch)
+* One single submission to HTCondor instead of multiple job submission ()
+* Ability to gracefully handle jobs with children (may require refactoring models)
```
MrCreosote, line 103: I'm not sure what this means - handling a retry of a batch parent job?

MrCreosote (commented):
This is the review for ADR #2

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -0,0 +1,118 @@
...
+* The retry will only continue if the status of the job to be retried is in [Status.terminated.value, Status.error.value]
+* If the job id points to a job that has already been retried, it will attempt to retry that job's `retry_parent` instead.
+* If the job id has never been retried, it becomes the `retry_parent` 
+* EE2 looks up the job versions and parameters, and then submits the job to be retried, incrementing the `retry_count`
+  of the job being retried, and the newly launched job gains a pointer to the `_PARENT_RETRY_JOB_ID`
+* The job is submitted and upon successful submission, notifies the `retry_parent` and notifies the `parent_job_id` that a new `child_job` has been added
+
+
+### Batch Behavior
+* If a job has the attribute of `batch_job=True` the retry will fail, since there is no method to re-run. This is a bug, as it doesn't fail gracefully.
```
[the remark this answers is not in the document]
bio-boris: It now fails gracefully, and says you can't re-run a batch job, instead of dying because it doesn't have a method

PR #368, opened 2021-04-28
DATAUP-423 Removed extra logging

bio-boris:
Remove UN-necessary logging

### DIFF
--- a/lib/execution_engine2/utils/Condor.py
+++ b/lib/execution_engine2/utils/Condor.py
@@ -311,8 +311,6 @@ class Condor:
             cancel_jobs = self.htcondor.Schedd().act(
                 action=self.htcondor.JobAction.Remove, job_spec=scheduler_ids
             )
-            self.logger.info(f"Cancel job message for {scheduler_ids} is")
-            self.logger.debug(f"{cancel_jobs}")
             return cancel_jobs
         except Exception:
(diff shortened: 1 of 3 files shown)


lib/execution_engine2/utils/Condor.py
```diff
@@ -197,18 +197,14 @@ def run_job(self, params: JobSubmissionParameters) -> SubmissionInfo:
...
         sub = self.htcondor.Submit(submit)
         try:
             schedd = self.htcondor.Schedd()
-            self.logger.debug(schedd)
-            self.logger.debug(submit)
-            self.logger.debug(os.getuid())
-            self.logger.debug(pwd.getpwuid(os.getuid()).pw_name)
-            self.logger.debug(submit)
             with schedd.transaction() as txn:
                 return SubmissionInfo(str(sub.queue(txn, 1)), sub, None)
```
MrCreosote, line 209: Out of scope for this PR but I think the 2nd argument to `SubmissionInfo` is supposed to be `submit` vs. `sub`

MrCreosote: Although maybe it should be nothing so the secrets aren't exposed outside of the Condor module

MrCreosote (approved):
LGTM

PR #372, opened 2021-04-29
SCT-3026 Upgrade mongo-engine

bio-boris:
"The id field is already used to implement the time_span filter, resulting it its usage in a regular field. This prevents the usage of id in a user-provided filter, which is used in a _raw_ filter."

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ maps==5.1.1
 MarkupSafe==1.1.1
 memory-profiler==0.55.0
 mock==3.0.5
-mongoengine==0.18.2
+mongoengine==0.23.0
 multidict==4.5.2
 nose==1.3.7
 packaging==20.4


MrCreosote:
I don't really understand the comment, which I assume is from the ME release notes. Does this affect us in some way? 

Would you mind updating the pipfile? If that's a problem no big deal; I can just update it on my next PR.

bio-boris:
@MrCreosote The info about why to upgrade this mongo engine is in https://kbase-jira.atlassian.net/browse/SCT-3026

I haven't really used pipfile before. Do you mind updating this PR or leaving a comment on how to do it? I see a piplock file too which seems to need updating too.

MrCreosote:
You can use https://docs.pipenv.org/ to update the pipfile and lock file. One quirk is that pipenv doesn't pin the version by default so I do that manually. Feel free to leave things as is and I'll update when this shows up in my local repo.

MrCreosote (approved):

PR #373, opened 2021-04-29
DATAUP-389 - Add integration tests for run_job_batch

MrCreosote:

What it says on the tin. Required a recompile of the client to add the batch run method, which is about 1000 lines of changes.

# Testing Instructions

# Dev Checklist:

- [?] My changes generate no new warnings - 3k warnings


lib/installed_clients/execution_engine2Client.py
```diff
@@ -102,117 +102,165 @@ def run_job(self, params, context=None):
         """
         Start a new job (long running method of service registered in ServiceRegistery).
         Such job runs Docker image for this service in script mode.
-        :param params: instance of type "RunJobParams" (method - service
```
MrCreosote, line 105: Recompiled the ee2 client, hence the 1000 lines of changes here

test/tests_for_integration/api_to_db_test.py
```diff
@@ -924,3 +933,335 @@ def _run_job_concierge_fail(
...
+            {
+                "+KB_MODULE_NAME": '"mod2"',
+                "+KB_FUNCTION_NAME": '"meth2"',
+                "+KB_APP_ID": '"mod2/app2"',
+                "+KB_APP_MODULE_NAME": '"mod2"',
+                "+KB_SOURCE_WS_OBJECTS": "",
+            }
+        )
+
+        assert sub_init.call_args_list == [call(expected_sub_1), call(expected_sub_2)]
```
bio-boris, line 1140: The tests look reasonable to me. Not sure what's going on with the mocked objects in here, but I wonder if knowing that the HTC Condor objects are supposed to be doing things with threads would help.

Also, are we OK with the tradition of very long tests?

bio-boris (approved):

test/tests_for_integration/api_to_db_test.py
MrCreosote: > I wonder if knowing that the HTC Condor objects are supposed to be doing things with threads would help.

I'm not 100% sure, but I don't think so. For some reason the mocking lib is including subcalls to the mocks in the list of calls for a particular mock, which I don't think can be due to threading, because the Condor code shouldn't be called at all since it's all mocked out.

MrCreosote: > Also, are we OK with the tradition of very long tests?

I mean, I prefer short tests, but in this case it's pretty hard to shorten it up without doing kind of pointless things like making methods for creating each argument and response set. A large portion of the length is due to large data structures that are being checked - the actual LOC is probably reasonable, but the test looks huge due to the inline data structures.

MrCreosote: Personally, I'd rather just have the data structures in line than have to jump around to various functions to see what they are *shrug*

MrCreosote: And finally, integration tests are often pretty hairy, which is one reason to keep them to a minimum

PR #376, opened 2021-04-30
DATAUP-424 remove slack calls

bio-boris:

* Removes messages/dependency on slack for run_job
* We could put it back in for run_job, but not for run_job_batch, but if we do, we should wrap in try/catch and also launch a thread or process and not wait for it, and not be blocked by it

### DIFF
--- a/build/templates/deploy.docker.cfg.templ
+++ b/build/templates/deploy.docker.cfg.templ
@@ -48,6 +48,11 @@ transfer_input_files = ../scripts/JobRunner.tgz
 debug = false
 
 #---------------------------------------------------------------------------------------#
+[concierge]
+request_cpus = 4
+request_memory = 23000M
+request_disk = 100GB
+#---------------------------------------------------------------------------------------#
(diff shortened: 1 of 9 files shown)


bio-boris:
tagging @kbase/du_devs in the reviewers randomly chooses 3 reviewers, pretty neat

MrCreosote:
Looks like tests are failing

MrCreosote (changes requested):
Changes look fine, but I'm guessing that EE2RunJob_test.py needs to be altered, as it checks for calls to slack as part of the test

MrCreosote:
IIRC there's 2 slack calls - one for create and one for update

bio-boris:
> IIRC there's 2 slack calls - one for create and one for update

Hmm I'm not seeing it... Am I missing something?

MrCreosote:
Never mind, I was thinking of kafka. Sorry

MrCreosote (approved):
LGTM

PR #380, opened 2021-05-03
DATAUP-389 - relax app_id requirements

MrCreosote:

There are some KBase processes (KBParallel and Narrative downloads,
perhaps more) where it doesn't make sense to provide an app_id but app_id is
currently a required field. This PR:
- makes app_id an optional field
- allows a . rather than a / as the app ID separator as Narrative downloads use
  the method ID as the app ID
- allows just the module name as the app ID as KBParallel sends that

Based on an analysis of the production EE2 DB, there aren't any other regular
cases of bad app IDs.

In the future, as processes stop submitting incorrect app_ids, we may be able to
tighten the restrictions on app IDs again.

# Testing Instructions

# Dev Checklist:

- [?] My changes generate no new warnings - 3k warnings


lib/execution_engine2/utils/application_info.py
```diff
@@ -2,15 +2,19 @@
 Contains information about KBase applications.
 """
 
+from typing import Union
 from execution_engine2.utils.arg_processing import check_string as _check_string
 from execution_engine2.exceptions import IncorrectParamsException
 
 
-def _get2part_string(s, sep, name):
+def _get2part_string(s, sep, name, err_pt1, err_pt2, desired_sep=None):
```
MrCreosote, line 10: `desired_sep` is so the error message encourages the user to use a `/` separator rather then reinforcing the use of the incorrect `.` separator.

lib/execution_engine2/utils/application_info.py
```diff
@@ -21,57 +25,79 @@ class AppInfo:
...
 
         method - the method name, e.g. kb_uploadmethods.import_reads_from_staging
-        app_id - the app name, either fully qualified (e.g.
-            kb_uploadmethods/import_fastq_interleaved_as_reads_from_staging or unqualified (e.g.
-            import_fastq_interleaved_as_reads_from_staging). If fully qualified, the module name
-            of the app (kb_uploadmethds in this example) must match the module name for the method.
+        app_id - the app name in the module/app_name format (e.g.
+            kb_uploadmethods/import_fastq_interleaved_as_reads_from_staging). For historical
+            reasons, this class will also accept only the module name or the module.app_name
+            format. In both cases the module name must match that given for the method argument.
```
bio-boris, line 42: I wonder if we should be storing "kbparallels", "concierge" or "cli ee2 call" or something in that field

test/tests_for_integration/api_to_db_test.py
```diff
@@ -448,8 +457,8 @@ def _get_condor_sub_for_rj_param_set(
             "+KB_PARENT_JOB_ID": f'"{parent_job_id}"',
             "+KB_MODULE_NAME": '"mod"',
             "+KB_FUNCTION_NAME": '"meth"',
-            "+KB_APP_ID": f'"{_APP}"',
-            "+KB_APP_MODULE_NAME": '"mod"',
+            "+KB_APP_ID": f'"{app_id}"' if app_id else "",
```
bio-boris, line 460: I wonder if an empty string without quotes works in the else condition. I think the simplest way to test it out is to submit the job and see what happens to it.

bio-boris (approved):
LGTM, lets get this up on CI

lib/execution_engine2/utils/application_info.py
MrCreosote: I don't think so - IMO each field should have a particular meaning if present. If I'm a user, I don't want a field to have more than one meaning that I have to figure out from the contents. If we want to have a field where we store the source of the job call, we should add another field, again IMO.

test/tests_for_integration/api_to_db_test.py
MrCreosote: If including quotes is the normal way of doing things, I'd rather just update the code.

MrCreosote:
I'm going to merge for now to get the next PR up, but if I should quote empty strings let's just do that - I'll make a quick PR with the fix.

MrCreosote:
For posterity, it's not clear whether the strings should be quoted as not, so we're going to try it as is.

PR #393, opened 2021-06-02
Update 003-Retry_endpoint_design.md

bio-boris:
Retry endpoint ADR


docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
-* Prevent multiple in-flight retries to prevent the user from wasting their own resources (and the queues resources)
+* Prevent multiple in-flight retries of the same original job to prevent the user from wasting their own resources (and the queues resources)
 * Add retry_count to retried jobs as well to aid in more book-keeping in a new field called `retry_number`
 * Non blocking job submission for submitting multiple jobs, possibly via using `run_job_batch` (requires refactor of run_job_batch)
 * One single submission to HTCondor instead of multiple job submissions
-* Ability to gracefully handle jobs with children
-* Ability to handle database consistentcy during retry failure
+* Ability to handle database consistency during retry failure. (see thread https://github.com/kbase/execution_engine2/pull/383#discussion_r638341940)
+* Add failure conditions in the run() method and see if any of those should be checked prior to starting jobs
+* Prevent retry if resolving `retry_parent` results in two or more of the same job_ids in a retry_jobs request (see thread https://github.com/kbase/execution_engine2/pull/383#discussion_r640907736)
```
MrCreosote, line 57: Or silently just run the one retry job and report the results to both positions in the list

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
 
 
 ### Questions
```
MrCreosote, line 60: I can't select the correct line, but on line 68 below:
I'm not sure what `same params` means here - the job parameters? I assume not because of course we should retry with the same params...

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -100,7 +103,6 @@ A: Not necessarily relevant to this endpoint, more of a run_job_batch endpoint q
 Priority descending
 * Non blocking job submission for submitting multiple jobs, possibly via using `run_job_batch` (requires refactor of run_job_batch)
 * One single submission to HTCondor instead of multiple job submission ()
-* Ability to gracefully handle jobs with children (may require refactoring models)
 * Prevent multiple in-flight retries to prevent the user from wasting their own resources (and the queues resources)
 * Add retry_count to retried jobs as well to aid in more book-keeping in a new field called `retry_number`
```
MrCreosote, line 107: I think that anything that might change the API needs to be top priority so the FE folks don't code against something that's going to change. I think the only issue that fits that description is adding the retry number to the child jobs if we decide to do that.

The remainder of the stuff is more loosy goosy but IMO dealing with potential data inconsistency should be near the top.

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
 * EE2 looks up the job versions and parameters, and then submits the job to be retried, incrementing the `retry_count`
   of the job being retried, and the newly launched job gains a pointer to the `_PARENT_RETRY_JOB_ID`
 * The job is submitted and upon successful submission, notifies the `retry_parent` and notifies the `parent_job_id` that a new `child_job` has been added
+* Jobs submitted by an administrator with a ResourceRequirements use resource requirements from the catalog / ee2 config exclusively
```
MrCreosote, line 38: I would clarify by ending with "exclusively on retry"

MrCreosote:
Hmm, looks like tests are failing for some reason?

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
+### Batch Jobs Behavior
+* Adds child_job_id to parent_job_id.child_job_ids
 
 ## Retry_job behavior
-* Blocking and single submit to HTCondor. It should be fine
+* Blocking and single submit to HTCondor. It should be fine as it returns relatively quickly
   
 ## Retry_jobs behavior
-* Submitting multiple jobs uses the `run_job` endpoint, and is blocking (NOT OK!)
+* Submitting multiple jobs uses the `run_job` endpoint, and is blocking. This can cause issues if the network drops, and makes the narrative not aware of the state of the retry. Submitting 100 jobs currently takes 12 seconds, and that is a lot of time for things to go wrong.
```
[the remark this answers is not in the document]
bio-boris: Yes. Run_job is the endpoint

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Added

docs/adrs/003-Retry_endpoint_design.md
bio-boris: done

bio-boris:
Tests are fixed in other pr

bio-boris:
> Hmm, looks like tests are failing for some reason?

Fixed in other PR

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Updated the wording "Identical params" for multiple jobs in the same retry

docs/adrs/003-Retry_endpoint_design.md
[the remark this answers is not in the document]
MrCreosote: I'm confused by this answer - the retry endpoint is `retry_jobs` surely? https://github.com/kbase/execution_engine2/blob/develop/execution_engine2.spec#L247

docs/adrs/003-Retry_endpoint_design.md
MrCreosote: :+1:

docs/adrs/003-Retry_endpoint_design.md
MrCreosote: So by `params` you mean the job parameters I'm assuming.

What would be the argument for disallowing re running jobs that have the same job parameters in the same retry call?

docs/adrs/003-Retry_endpoint_design.md
MrCreosote: :+1:

MrCreosote:
Now mergable is complaining - looks like the PR template didn't get filled in

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -1,11 +1,11 @@
...
 
 Date: 2021-05-19
 
 
 ## Motivation for the Endpoint:
 
-The current requirement for the Batch/Bulk UI is to be able to retry jobs that have either "errored" out, or were cancelled.
+The current requirement for the Batch/Bulk UI is to be able to retry jobs that have either "errored" out, or were terminated.
 The UI allows you to retry either single jobs, or multiple jobs, and saves you from having to cancel and resubmit each job individually,
 which is not really possibly with the UI anyway.
```
briehl, line 10: *possible

And it IS possible, just not implemented, and would be a huge pain. Selecting a number of jobs and hitting Retry is the way to go.

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
 * The job is submitted and upon successful submission, notifies the `retry_parent` and notifies the `parent_job_id` that a new `child_job` has been added
+* Jobs submitted by an administrator with a ResourceRequirements use resource requirements from the catalog / ee2 config exclusively on retry
 
-
-### Batch Behavior
-* If a job has the attribute of `batch_job=True` the retry will fail, since there is no method to re-run. This is a bug, as it doesn't fail gracefully.
-* If a job has the attribute of `batch_job=True`, but is actually a child job, the parent will be notified of this new retried job
-* Multiple in-flight retries are allowed.
+### Batch Jobs Behavior
+* Adds child_job_id to parent_job_id.child_job_ids
```
briehl, line 41: If I'm reading this right, the parent job just gets a new child job id for the retried job. I may not have followed, is there a way to get the original job from the retried job id?

docs/adrs/003-Retry_endpoint_design.md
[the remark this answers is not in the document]
bio-boris: Yes, the retry endpoint is used for retrying jobs. The retry endpoint doesn't submit jobs though, it uses the path taken by `run_job` (aka run_job function, as opposed to using the run_jobs function) in order to create a record in mongo and submit to htcondor though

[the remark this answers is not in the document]
briehl: Yeah, that looks like a typo - should read "Submitting multiple jobs uses the `retry_jobs` endpoint", right?

docs/adrs/003-Retry_endpoint_design.md
bio-boris: The argument is you are running the same job, so in most cases it's a waste of a job, why submit the same thing twice?

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Yes, the original job is available via `retry_parent` field

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -1,11 +1,11 @@
...
 
 Date: 2021-05-19
 
 
 ## Motivation for the Endpoint:
 
-The current requirement for the Batch/Bulk UI is to be able to retry jobs that have either "errored" out, or were cancelled.
+The current requirement for the Batch/Bulk UI is to be able to retry jobs that have either "errored" out, or were terminated.
 The UI allows you to retry either single jobs, or multiple jobs, and saves you from having to cancel and resubmit each job individually,
 which is not really possibly with the UI anyway.
```
ialarmedalien, line 11: comment for lines 13 - 20: can we change this to being quoted text, rather than being styled as code?

See https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-text

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Ok I will change the wording to "Which is not implemented in the UI"

docs/adrs/003-Retry_endpoint_design.md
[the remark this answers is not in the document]
briehl: Hopefully the making the narrative aware of the state of the retry will be mitigated by the narrative backend. It just blocks on the call anyway, with a minute timeout.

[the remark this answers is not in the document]
briehl: As long as the user doesn't kill the kernel at that time, all should be well. Of course, if it were me, and it looked frozen for more than a couple minutes, I'd probably restart.

docs/adrs/003-Retry_endpoint_design.md
ialarmedalien: Comment for line 26: rephrase to

> The endpoint takes a job or list of job ids and attempts to resubmit them to the queue, using the same parameters and version of the app as the original job.

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
 * EE2 looks up the job versions and parameters, and then submits the job to be retried, incrementing the `retry_count`
```
ialarmedalien, line 35: job versions? Should that be app version?

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
   of the job being retried, and the newly launched job gains a pointer to the `_PARENT_RETRY_JOB_ID`
```
ialarmedalien, line 36: does `_PARENT_RETRY_JOB_ID` refer to the name of the field in mongodb? The terminology should be consistent, so this should refer to assigning the `retry_parent` of the newly-launched job, rather than pointers and db field names.

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
 * The job is submitted and upon successful submission, notifies the `retry_parent` and notifies the `parent_job_id` that a new `child_job` has been added
```
ialarmedalien, line 37: the job doesn't notify anything itself -- ee2 updates the parent job with the new child job ID and updates the retry count of the job that has been retried

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -35,26 +35,26 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
+* Jobs submitted by an administrator with a ResourceRequirements use resource requirements from the catalog / ee2 config exclusively on retry
```
ialarmedalien, line 38: what does "exclusively" refer to in this sentence?

docs/adrs/003-Retry_endpoint_design.md
ialarmedalien: can you ensure that you quote field names consistently throughout the ADR?

docs/adrs/003-Retry_endpoint_design.md
ialarmedalien: > dealing with potential data inconsistency should be near the top.

💯

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -64,8 +64,8 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
 #### Q: should the number of retries of a job be limited, and if so, where? e.g. a max_retries field in the parent job? wait and see whether people attempt to rerun jobs that have already failed nine zillion times?
 A: Unknown TBD
 
-#### Q: Preventing the same params from being re-run
-A: We have decided to allow it
+#### Q: Preventing the identical params from being re-run within a retry_jobs request
```
ialarmedalien, line 67: Turn this into a question. I'm not sure which of the following interpretations you mean, but either

> How do we prevent the same job from being rerun more than once within a `retry_jobs` request?

or 

> How do we prevent jobs with identical parameters from being rerun more than once within a `retry_jobs` request?

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -64,8 +64,8 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
+A: We have decided to allow multiple jobs with the same params to be re-run in the same retry_jobs request.
```
ialarmedalien, line 68: when did we decide this? is it a temporary decision until the behaviour can be patched to eliminate duplicates?

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -64,8 +64,8 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
 
 #### Q: Finding the most recent run of the job: I would very much like to avoid anything involving iterating over a chain of jobs before you can find the most recent run or the original run -- we can come up with better data structures than that!
 A: Unknown TBD, maybe the frontend does it?
```
ialarmedalien, line 71: The main concern I had when writing that question was about whether a job with a `retry_parent` could itself be a retry parent. That won't happen, so the bit about the data structures isn't relevant. I'm totally fine with deleting this question, but if it would be of use, it can be edited to be something like

Q: How do we find the most recent retry of a job?

A: The client using the ee2 API would have to figure it out using the `retry_parent` and job creation date fields.

docs/adrs/003-Retry_endpoint_design.md
bio-boris: We decided this in our last engineering meeting

docs/adrs/003-Retry_endpoint_design.md
bio-boris: ok edited

docs/adrs/003-Retry_endpoint_design.md
bio-boris: done

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Updated to say Jobs submitted by an administrator with a ResourceRequirements use resource requirements from the catalog / ee2 config exclusively on retry. There is no way to specify ResourceRequirements with a retry at the moment.

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Updated The job is submitted and upon successful submission, the child job adds the field retry_parent and notifies the parent_job_id that a new child_job has been added by appending itself to the parent_job.child_jobs[] field

docs/adrs/003-Retry_endpoint_design.md
bio-boris: updated

docs/adrs/003-Retry_endpoint_design.md
bio-boris: updated

docs/adrs/003-Retry_endpoint_design.md
bio-boris: done

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Alright, I think I got them all

docs/adrs/003-Retry_endpoint_design.md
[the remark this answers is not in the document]
MrCreosote: >  The retry endpoint doesn't submit jobs though, it uses the path taken by run_job
But it doesn't, which is why I'm confused. The call stack is
```
retry_jobs -> retry_multiple -> _retry -> run -> _run
```
the call there calls methods on `Condor`, so `run_job` is never in the call stack

docs/adrs/003-Retry_endpoint_design.md
MrCreosote: Ok, so that means the user has run the same job twice previously, either by `run_job` or `run_jobs`, and the user is now trying to retry them both. In order to avoid this error the user would have to pull the job params and compare them with each other to be sure they're not the same, which seems pretty onerous.

Also presumably users may want to run the same job again if the app version changes.

I don't think any changes are necessary just thinking out loud.

docs/adrs/003-Retry_endpoint_design.md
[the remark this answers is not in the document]
bio-boris: The path by run_job to submit to condor begins with `run(params=params, as_admin=as_admin)`
The path by retry to submit condor begins with `retry_job_id = self.run(params=run_job_params, as_admin=as_admin)`

They aren't calling run_job_batch, for example.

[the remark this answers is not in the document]
MrCreosote: > I never mentioned a call stack, and I clarified in the comment that its using the path taken by run_job.. What do you want to sentence to say instead?

I don't currently want it to say anything in particular because I don't really understand the point that it's trying to make, and when I read it it looks incorrect, e.g.:
```
## Retry_jobs behavior
Submitting multiple jobs uses the `run_job` endpoint, and is blocking
```
That makes it sound like `retry_jobs` calls `run_job` under the hood, which it doesn't - that's what I was talking about when I posted the stack.

How about I back up and just ask this - what is the message that the reader of that sentence should get?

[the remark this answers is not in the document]
bio-boris: The message of the reader of the sentence should get is 

Job submission for multiple jobs uses the same business logic as the run_job endpoint and thus is currently blocking. run_jobs_batch is not used.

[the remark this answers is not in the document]
MrCreosote: `run_jobs_batch` is blocking too, right? That seems to imply that it isn't.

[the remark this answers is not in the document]
MrCreosote: Ok, I would suggest something like
```
Submitting multiple jobs for retry serially calls the same code path used for running a single job and blocks until all jobs have been submitted to the condor queue. This can cause issues if the network drops, and makes the narrative not aware of the state of the retry. Submitting 100 jobs currently takes 12 seconds, and that is a lot of time for things to go wrong. 
```
Does that sound right?

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -64,26 +64,29 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
 
-#### Q: Preventing the same params from being re-run
-A: We have decided to allow it
+#### Q: How do we prevent jobs with identical parameters from being rerun more than once within a retry_jobs request?
+A: We have decided to allow multiple jobs with the same params to be re-run in the same `retry_jobs` request.
 
-#### Q: Finding the most recent run of the job: I would very much like to avoid anything involving iterating over a chain of jobs before you can find the most recent run or the original run -- we can come up with better data structures than that!
-A: Unknown TBD, maybe the frontend does it?
+#### Q: How do we find the most recent retry of a job?
+A: The client using the ee2 API would have to figure it out using the `retry_parent` and job creation date fields. Unless we added
```
ialarmedalien, line 71: unfinished sentence

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -64,26 +64,29 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
-A: We have decided to allow it
+#### Q: How do we prevent jobs with identical parameters from being rerun more than once within a retry_jobs request?
+A: We have decided to allow multiple jobs with the same params to be re-run in the same `retry_jobs` request.
 
-#### Q: Finding the most recent run of the job: I would very much like to avoid anything involving iterating over a chain of jobs before you can find the most recent run or the original run -- we can come up with better data structures than that!
-A: Unknown TBD, maybe the frontend does it?
+#### Q: How do we find the most recent retry of a job?
+A: The client using the ee2 API would have to figure it out using the `retry_parent` and job creation date fields. Unless we added 
 
 #### Q: It might be best to always submit a git commit for the module, maybe?
```
ialarmedalien, line 73: Reword to:

Q. How do we ensure that the correct version of an app is always run?

A. The git commit for an app is the most accurate way to specify the version. Ideally any clients using the ee2 API would supply this information.

bio-boris: The clients mostly use "dev/beta/release" within the app runs, but this was actually meant to talk about how the narrative sends dev/beta/release instead of git commits, so fixed it up to reflect it.

docs/adrs/003-Retry_endpoint_design.md
bio-boris: Updated it

docs/adrs/003-Retry_endpoint_design.md
bio-boris: @ialarmedalien I think the ADR already reflects this, so not sure what you mean. It says

`Multiple in-flight retries are allowed` as the current behavior and
`Prevent multiple in-flight retries of the same original job to prevent the user from wasting their own resources (and the queues resources)` is the desired behavior

bio-boris: I would agree with you that it is a parameter, but my comment was  off of the models https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/db/models/models.py#L156 where you can see the Params are job params, and Source_ws_object is just a job input.

docs/adrs/003-Retry_endpoint_design.md
bio-boris: done

docs/adrs/003-Retry_endpoint_design.md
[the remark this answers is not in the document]
bio-boris: Updated it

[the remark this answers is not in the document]
MrCreosote: :+1:

docs/adrs/003-Retry_endpoint_design.md
ialarmedalien: OK -- I think this is all resolved, then. 🎉

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -32,29 +31,38 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
-* Prevent multiple in-flight retries to prevent the user from wasting their own resources (and the queues resources)
-* Add retry_count to retried jobs as well to aid in more book-keeping in a new field called `retry_number`
-* Non blocking job submission for submitting multiple jobs, possibly via using `run_job_batch` (requires refactor of run_job_batch)
-* One single submission to HTCondor instead of multiple job submissions
-* Ability to gracefully handle jobs with children
-* Ability to handle database consistentcy during retry failure
+* Create a retry_jobs field, and a toggle to have a list of retried jobs and to ensure DB consistency
+* Prevent multiple in-flight retries of the same original job to prevent the user from wasting their own resources (and the queues resources)
+* Ability to handle database consistency during retry failure. (see thread https://github.com/kbase/execution_engine2/pull/383#discussion_r638341940)
+* Add failure conditions in the `run()` method and see if any of those should be checked prior to starting jobs
```
ialarmedalien, line 60: Fail quicker if the job is going to fail

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -32,58 +31,75 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
+* * One single submission to HTCondor instead of multiple job submissions
 
-### Questions
+#### Won't do
+* Prevent multiple in-flight retries of the same original job to prevent the user from wasting their own resources (and the queues resources)
+* Add retry_number field
 
-#To Be Answered
+## New priority
+* Create a retry_jobs field, and expose list in api, and a T/F completeness toggle
```
MrCreosote, line 69: From what I recall the items were:
1. add a `completed_processing` (or whatever) boolean flag on the retry child job that starts out `False`
2. the retry child job should always have a retry parent field
3. the retry child job should have a batch parent field if it's part of a batch job
4. after creating the retry job, its ID will be added to a list of retry children on the retry parent using `$addToSet`
5. after creating the retry job, its ID will be added to a list of batch children on the batch parent using `$addToSet` if applicable
6. after that, set `completed_processing` to `True`

This means data integrity can be checked and corrected by scanning for jobs in an end state (complete / error / terminated) and doing steps 4-6 for each job.

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -32,58 +31,75 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
+* Non blocking job submission / (Possibly htcondor submit)
+* Add failure conditions in run method
+* Add thread to perform actions based on toggle
```
MrCreosote, line 72: e.g. `complete_processing` above

docs/adrs/003-Retry_endpoint_design.md
```diff
@@ -32,58 +31,75 @@ The endpoint takes a job or list of job ids and then attempts to resubmit them t
...
-### Questions
+#### Won't do
+* Prevent multiple in-flight retries of the same original job to prevent the user from wasting their own resources (and the queues resources)
+* Add retry_number field
 
-#To Be Answered
+## New priority
+* Create a retry_jobs field, and expose list in api, and a T/F completeness toggle
+* Non blocking job submission / (Possibly htcondor submit)
+* Add failure conditions in run method
```
MrCreosote, line 71: More specifically, look for checks that happen in the run() method and see if they're handled sanely, esp. if we thread starting jobs.

E.g.
- look for quick checks that could be moved prior to creating database records, and for which we should fail the entire submission.
- for checks where it doesn't make sense or is infeasible to perform prior to creating database records and potentially starting a thread to complete processing asynchronously, ensure that the database records are updated correctly if they fail. This applies to exception handling as well.

MrCreosote (commented):
Added some more detail from the engineering check in while it was in my head

PR #397, opened 2021-06-15
DATAUP-461 Refactor Retry

bio-boris:

Summary:
In order to ensure DB consistency, we need to rework the retry mechanism

    Remove retry_count field from record
    Add retry_ids field to record
    Change check_job/check_jobs to return “retry_count” in addition to “retry_ids”
    Change retry to return “retry_ids”
    Add a toggle field to ensure database is consistent in terms of updating the parent_retry and parent_job_id  

# Testing Instructions
* Deployed on CI
* Delete all records with retries in CI db, and prod db.

```
db.getCollection('ee2_jobs').update({'retry_count' : { $exists: true }}, { $set: {"retry_ids": [] }}, {upsert:false,
                          multi:true})  

db.getCollection('ee2_jobs').update({'retry_count' : { $exists: true }}, { $unset: {"retry_count" : ""}}, {upsert:false,
                          multi:true})  

db.getCollection('ee2_jobs').find({'retry_count' : { $exists: true }})
```

# Dev Checklist:

### DIFF
--- a/execution_engine2.spec
+++ b/execution_engine2.spec
@@ -459,7 +459,6 @@
         retry_count - int - generated field based on length of retry_ids
         retry_ids - list - list of jobs that are retried based off of this job
         retry_parent - str - job_id of the parent this retry is based off of. Not available on a retry_parent itself
-        retry_saved_toggle - bool - Marked true when all retry steps/txns have completed
 
         parent_job_id - str - job_id taken from job_input.parent_job_id
         batch_job - bool - whether or not this is a batch parent container
--- a/execution_engine2.spec
+++ b/execution_engine2.spec
@@ -428,6 +428,7 @@
         wsid - int - optional id of the workspace where the job is bound
         authstrat - string - what strategy used to authenticate the job
         job_input - object - inputs to the job (from the run_job call)  ## TODO - verify
+        job_output - object - outputs from the job (from the run_job call) ## TODO - verify
         updated - int - timestamp since epoch in milliseconds of the last time the status was updated
         running - int - timestamp since epoch in milliseconds of when it entered the running state
         created - int - timestamp since epoch in milliseconds when the job was created
@@ -452,18 +453,35 @@
         errormsg - string - message (e.g. stacktrace) accompanying an errored job
         error - object - the JSON-RPC error package that accompanies the error code and message
 
-            terminated_code - int - internal reason why a job was terminated, one of:
-                0 - user cancellation
-                1 - admin cancellation
-                2 - terminated by some automatic process
+        #TODO, add these to the structure?
+        condor_job_ads - dict - condor related job information
 
-            @optional error
-            @optional error_code
-            @optional errormsg
-            @optional terminated_code
-            @optional estimating
-            @optional running
-            @optional finished
+        retry_count - int - generated field based on length of retry_ids
+        retry_ids - list - list of jobs that are retried based off of this job
+        retry_parent - str - job_id of the parent this retry is based off of. Not available on a retry_parent itself
+        retry_saved_toggle - bool - Marked true when all retry steps/txns have completed
+
+        parent_job_id - str - job_id taken from job_input.parent_job_id
+        batch_job - bool - whether or not this is a batch parent container
+        child_jobs - array - Only parent container should have child job ids
+
+        scheduler_type - str - scheduler, such as awe or condor
+        scheduler_id - str - scheduler generated id
+        scheduler_estimator_id - str - id for the job spawned for estimation
+
+
+        terminated_code - int - internal reason why a job was terminated, one of:
+            0 - user cancellation
+            1 - admin cancellation
+            2 - terminated by some automatic process
+
+        @optional error
+        @optional error_code
+        @optional errormsg
+        @optional terminated_code
+        @optional estimating
+        @optional running
+        @optional finished
         */
 
 
@@ -484,6 +502,7 @@
             int error_code;
             string errormsg;
             int terminated_code;
+
         } JobState;
 
         /*
--- a/lib/execution_engine2/sdk/EE2Runjob.py
+++ b/lib/execution_engine2/sdk/EE2Runjob.py
@@ -572,7 +572,7 @@ class EE2RunJob:
         # 1) Notify the parent container that it has a new child..
         if parent_job:
             try:
-                parent_job.modify(push__child_jobs=retry_job_id)
+                parent_job.modify(add_to_set__child_jobs=retry_job_id)
             except Exception as e:
                 self._db_update_failure(
                     job_that_failed_operation=str(parent_job.id),
@@ -582,7 +582,7 @@ class EE2RunJob:
 
         # 2) Notify the retry_parent that it has been retried by adding a retry id
         try:
-            job.modify(push__retry_ids=retry_job_id)
+            job.modify(add_to_set__retry_ids=retry_job_id)
         except Exception as e:
             self._db_update_failure(
                 job_that_failed_operation=str(job.id),
--- a/lib/execution_engine2/sdk/EE2Runjob.py
+++ b/lib/execution_engine2/sdk/EE2Runjob.py
@@ -55,6 +55,7 @@ _METHOD = "method"
 _APP_ID = "app_id"
 _PARENT_JOB_ID = "parent_job_id"
 _PARENT_RETRY_JOB_ID = "retry_parent"
+_RETRY_IDS = "retry_ids"
 _WORKSPACE_ID = "wsid"
 _SOURCE_WS_OBJECTS = "source_ws_objects"
 _SERVICE_VER = "service_ver"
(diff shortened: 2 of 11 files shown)


lib/execution_engine2/sdk/EE2Status.py
```diff
@@ -472,7 +472,9 @@ def check_jobs(
             else:
                 mongo_rec = job.to_mongo().to_dict()
                 del mongo_rec["_id"]
+                mongo_rec["retry_count"] = len(job["retry_ids"])
                 mongo_rec["job_id"] = str(job.id)
+                mongo_rec["parent_job_id"] = job.job_input.parent_job_id
```
bio-boris, line 477: request by @ialarmedalien

lib/execution_engine2/sdk/EE2Status.py
```diff
@@ -472,7 +472,9 @@ def check_jobs(
             else:
                 mongo_rec = job.to_mongo().to_dict()
                 del mongo_rec["_id"]
+                mongo_rec["retry_count"] = len(job["retry_ids"])
```
bio-boris, line 475: add a returned field instead of actually adding it to the model... oh yea need to update documentation?

execution_engine2.spec
```diff
@@ -452,18 +453,35 @@
...
-            @optional error_code
-            @optional errormsg
-            @optional terminated_code
-            @optional estimating
-            @optional running
-            @optional finished
+        retry_count - int - generated field based on length of retry_ids
+        retry_ids - list - list of jobs that are retried based off of this job
+        retry_parent - str - job_id of the parent this retry is based off of. Not available on a retry_parent itself
+        retry_saved_toggle - bool - Marked true when all retry steps/txns have completed
```
ialarmedalien, line 462: @n1mus additional fields for retried jobs

MrCreosote, line 462: Hmm, I was thinking `retry_saved_toggle` would not be exposed in the API - I'm still not sure why you'd want to do that.

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -579,15 +580,19 @@ def _retry(self, job_id: str, job: Job, parent_job: Job, as_admin: bool = False)
...
+            job.modify(push__retry_ids=retry_job_id)
         except Exception as e:
             self._db_update_failure(
                 job_that_failed_operation=str(job.id),
                 job_to_abort=retry_job_id,
                 exception=e,
             )
+        # If the retry_ids is updated and if present, the child_jobs, is updated, set toggle to true
+        retry_job = self.sdkmr.get_mongo_util().get_job(job_id=retry_job_id)
+        retry_job.retry_saved_toggle = True
```
MrCreosote, line 594: It seems safer to set just this field rather than saving the entire job in case there's a state change from another part of the code between `get_job` and `save_job`. Especially if we thread retries - the jobs could be `run` and then the state change from `created` to `queued` could collide with this modification.

test/tests_for_integration/api_to_db_test.py
```diff
@@ -97,6 +97,11 @@
 _MOD = "mod.meth"
 _APP = "mod/app"
 
+# Note (For pycharm, ensure mongo entry is pointing to localhost and uncomment the following)
+from test.utils_shared.test_utils import bootstrap
+
+bootstrap()
```
MrCreosote, line 103: From the comment above it sounds like the 2 lines above should be commented?

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
```diff
@@ -7,6 +7,7 @@
 from unittest.mock import patch
 
 import requests_mock
+from bson import ObjectId
```
MrCreosote, line 10: Usused

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
```diff
@@ -407,10 +413,13 @@ def test_retry_job(self, rq_mock, condor_mock):
         self.check_retry_job_state(parent_job_id, retry_from_original_again)
 
         for job in [original_job, retried_job, retried_job2, retried_job3]:
+            j = Job.objects.get(id=job["job_id"])
             if job == original_job:
                 assert original_job["retry_count"] == 3
+                assert not j.retry_saved_toggle
```
MrCreosote, line 419: Could you check here that the original job's `retry_ids` field contains all the children?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -579,15 +580,19 @@ def _retry(self, job_id: str, job: Job, parent_job: Job, as_admin: bool = False)
                     exception=e,
                 )
 
-        # 2) Notify the retry_parent that it has been retried
+        # 2) Notify the retry_parent that it has been retried by adding a retry id
         try:
-            job.modify(inc__retry_count=1)
+            job.modify(push__retry_ids=retry_job_id)
```
MrCreosote, line 585: How difficult would it be to add a test where this fails, showing that the retry toggle is still false? That's a key piece of behavior. You might need to add a SDKMR.modify_job() method so the modify can be mocked.

Maybe also failing on the parent job update as well.

MrCreosote (changes requested):
Looks pretty straightforward, just a few things.

execution_engine2.spec
bio-boris: Well, if we wanted to run a script that reaped via the API, or wanted to see all jobs that had failed during a recent outage by looking for that toggle, that would be useful. Otherwise we would have to specifically hide it from the check_job call

test/tests_for_integration/api_to_db_test.py
bio-boris: Removed the comment

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
bio-boris: done

bio-boris: removed

execution_engine2.spec
bio-boris: Or the frontend can display an error if it sees that field

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: On second thought a mock test wouldn't really be useful here. It'd need to be an integration test with mongo and I don't see a way to do that test cleanly.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: good catch, changed

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
bio-boris: done

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -579,15 +580,25 @@ def _retry(self, job_id: str, job: Job, parent_job: Job, as_admin: bool = False)
...
```
MrCreosote, line 585: I missed this in my prior review - this `$push` and the batch parent modification `$push` above need to be `$addToSet` to avoid duplicate IDs in the list

MrCreosote (changes requested):

execution_engine2.spec
MrCreosote: My thinking was that if a job was fetched from the DB where that toggle was `False` the server would just run a fix up routine to finish the update (at some point, not in this PR). If that's the case the toggle should never show any state other than `True` so there's no point in exposing it.

test/tests_for_integration/api_to_db_test.py
MrCreosote: :+1:

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
MrCreosote: :+1:

lib/execution_engine2/sdk/EE2Runjob.py

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
MrCreosote: :+1:

execution_engine2.spec
bio-boris: So if I ran a check_jobs_date_range and it returned 10000 jobs, and 500 of them had failed retries, would it wait for 500 of them to run the routine? Or would it just assume they are true and not wait for the routine to run? What if the routine failed for some of the jobs, would the whole api call fail?

MrCreosote: > and 500 of them had failed retries, would it wait for 500 of them to run the routine?
That's what I would do, because this failure mode is likely going to be very rare and jobs with the toggle left as `False` are going to be infrequent.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: this is done

execution_engine2.spec
bio-boris: this is done

lib/execution_engine2/sdk/EE2Runjob.py

execution_engine2.spec
MrCreosote: :+1:

MrCreosote (approved):
LGTM

PR #399, opened 2021-06-18
Refactor to run multiple jobs

bio-boris:

* prepping for batch running throughout the code base, starting with the preflight checks

- the first pass is to make running jobs preflight checks run from a list of jobs.
- Then the second pass is to make the jobs all return a series of docs to save into mongo with a bulk operation.
- Then the next pass is to make them all get submitted to condor with a bulk operation. 
- Then the next pass is to make this operation done via a thread. 
- Then the next pass is to make sure the retry_multiple is submitting all jobs via this bulk submit. 
- Then the next pass is to make sure that batch submits via the bulk submit. 

# Testing Instructions
- [ ] Changes available on CI

# Dev Checklist:

- [ ] My code follows the guidelines at https://sites.google.com/truss.works/kbasetruss/data-upload-project/development
- [ ] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the Github Actions build passes)


lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -224,7 +225,7 @@ def _finish_created_job(
             error=f"{exception}",
         )
 
-    def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParameters:
+    def _prepare_to_run(self, params) -> JobSubmissionParameters:
```
bio-boris, line 228: removed unused param

lib/execution_engine2/execution_engine2Impl.py
```diff
@@ -30,7 +30,7 @@ class execution_engine2:
     ######################################### noqa
     VERSION = "0.0.5"
     GIT_URL = "git@github.com:kbase/execution_engine2.git"
-    GIT_COMMIT_HASH = "8b6f4e1917dbdfa374e6f22b1f2adbe7eca5a24c"
+    GIT_COMMIT_HASH = "75d8824c809f3156760a4650a601d3d18df075dd"
```
bio-boris, line 33: Just autogenerated stuff here

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -691,32 +708,81 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+        :param concierge_params: a single concierge param dict to over-write requirements for all jobs
+        :param as_admin: as_admin checked previously, not verified here
+        :return:
         """
-        if as_admin:
-            self.sdkmr.check_as_admin(requested_perm=JobPermissions.WRITE)
-        else:
-            self._check_workspace_permissions(params.get(_WORKSPACE_ID))
-
+        begin_hjr = time.time()
```
bio-boris, line 724: keeping these in for now for deployment

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -190,16 +192,23 @@ def _check_workspace_permissions(self, wsid):
...
-            self.logger.debug(
-                f"User {self.sdkmr.user_id} doesn't have permission to run jobs in workspace {bad_ws}."
-            )
-            raise PermissionError(
-                f"User {self.sdkmr.user_id} doesn't have permission to run jobs in workspace {bad_ws}."
+    def _check_workspace_permissions_list(self, wsids) -> None:
+        """
+        Raise exception if you don't have permissions to run a job in any of these workspaces
+        """
+        unique_not_none_wsids = [wsid for wsid in list(set(wsids)) if wsid]
```
MrCreosote, line 199: I don't think you need the `list` before the `set`, `set` is iterable

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -338,35 +346,43 @@ def _run_batch(self, parent_job: Job, params):
...
 
-        parent_job = self._create_parent_job(wsid=wsid, meta=meta)
+    def run_batch(
+        self, params: List[Dict], batch_params: Dict, as_admin: bool = False
+    ) -> Dict[str, List[str]]:
+        """
+        :param params: List of RunJobParams (See Spec File)
+        :param batch_params: List of Batch Params, such as wsid (See Spec file)
+        :param as_admin: Allows you to run jobs in other people's workspaces
+        :return: A list of condor job ids or a failure notification
```
MrCreosote, line 379: These are ee2 job IDs not condor job IDs, right? And it's a dict not a list

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -691,32 +709,81 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
-        self, params=None, as_admin=False, concierge_params: Dict = None
-    ) -> Optional[str]:
+    def _handle_job_requirements(
+        self, params: List[Dict], concierge_params: Dict = None, as_admin: bool = False
+    ) -> None:
         """
+        Modifies params in place with job requirements
+
         :param params: SpecialRunJobParamsParams object (See spec file)
         :param params: RunJobParams object (See spec file)
```
MrCreosote, line 719: Duplicate conflicting definitions of `params`

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -338,35 +346,43 @@ def _run_batch(self, parent_job: Job, params):
...
         if type(params) != list:
             raise IncorrectParamsException("params must be a list")
-        wsid = batch_params.get(_WORKSPACE_ID)
-        meta = batch_params.get(_META)
 
+        for item in params:
+            if not item:
+                raise MissingRunJobParamsForBatchException()
+
+    def _preflight_batch(self, params: List[Dict], as_admin=False, parent_wsid=None):
```
MrCreosote, line 358: I think _preflight_batch and _preflight could be merged into one method.

MrCreosote (changes requested):
Looks reasonable other than I feel pretty strongly that there's no need for a separate preflight batch method. It looks to me if a parent wsid param is added to the regular preflight method that would work for batch as well and would DRY up duplicate code paths.

I haven't gotten deep into the tests yet as that change might cause some reshuffling.

lib/execution_engine2/exceptions.py
```diff
@@ -1,19 +1,37 @@
...
-    pass
+    def __init__(self, msg=None, *args, **kwargs):
+        super().__init__(msg or self.__doc__, *args, **kwargs)
 
 
 class ExecutionEngineException(Exception):
-    pass
+    """
+    Base Class for ee2 exceptions
+    Subclass exceptions use docstring as default message
```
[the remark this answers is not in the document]
bio-boris: Yea sounds good

lib/execution_engine2/exceptions.py
```diff
@@ -1,19 +1,37 @@
...
 
+class MissingRunJobParamsForBatchException(ExecutionEngineValueError):
+    """Provided an empty (RunJobParams) parameter mapping to run_batch params"""
+
+    pass
+
+
 class MissingRunJobParamsException(ExecutionEngineValueError):
-    """Missing a required run_job_parameter"""
+    """Provided an empty (RunJobParams) parameter mapping to run_batch params"""
```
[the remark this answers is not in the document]
bio-boris: updated

lib/execution_engine2/exceptions.py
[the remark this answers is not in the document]
bio-boris: updated

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: removed list(set()) in favor of set()

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Fixed

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Yea I was wondering what that was, so I left it.. Deleting it now

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -6,15 +6,21 @@
 
 from logging import Logger
 from typing import List, Dict, Any
-from unittest.mock import create_autospec, call
+from unittest.mock import create_autospec, call, Mock
```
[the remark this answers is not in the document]
bio-boris: done

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -267,6 +273,52 @@ def _create_reqs_dict(
...
+    params2["wsid"] = "1234"
+    rj2._preflight_batch([params, params2])
+    rj2._check_workspace_permissions_list.assert_called_once_with(
+        [params.get("wsid"), params2.get("wsid")]
+    )
+    rj2._handle_job_requirements.assert_called_once_with(
+        as_admin=False, params=[params, params2]
+    )
+    rj2._check_job_arguments.assert_called_once_with(
+        [params, params2], has_parent_job=True
```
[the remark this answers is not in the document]
bio-boris: Darn, and I thought I was finally writing better tests by using assert_called_once_with to make sure that the logic was being called, and if something in that function was deleted, this test would break. What would you recommend instead?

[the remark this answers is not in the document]
MrCreosote: The thing is I *should* be able to rename / delete / modify internal methods / variables without having the tests break as long as it doesn't change the external behavior. The external behavior is what we care about, not the implementation details. For EE2RunJob, I don't care if `_preflight` is called as long as the job requirements are saved in mongo and passed to Condor.

As such, tests should ideally always be against the public API of the code unit and any mocks that need to be provided to the code unit for it to run. My general test rules go like this:

1) Test one method or function call per test (except for trivial getters, I'll usually test them all in one test)
1a) Unless the method changes class state - then you may need to check the state as well.
2) Create mocks for any other code units that the code unit under test requires
3) Set up the expected returns for the mocks (these should be public methods that you expect the code unit under test to call)
4) Pass the mocks into the constructor / method / function as necessary.
5) assert that the return value of the method, if any, is correct.
6) assert that calls to the mocks (again, public API) are correct.

That's what the tests in this module do, so given there's no new behavior or new methods in this PR, I think you should just have to adjust the tests for the minor changes to workspace checking, although I haven't check that in detail.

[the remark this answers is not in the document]
MrCreosote: Actually you may have added some more error checking to `run_one_job` now I think about it.

PR #402, opened 2021-07-01
DATAUP-528 Refactor preflight and wsids

bio-boris:
* prepping for batch running throughout the code base, starting with the preflight checks
* removing support for wsid in batch jobs outside of batchjobparams
* this pr in favor of #399 

- the first pass is to make running jobs preflight checks run from a list of jobs.
- Then the second pass is to make the jobs all return a series of docs to save into mongo with a bulk operation.
- Then the next pass is to make them all get submitted to condor with a bulk operation. 
- Then the next pass is to make this operation done via a thread. 
- Then the next pass is to make sure the retry_multiple is submitting all jobs via this bulk submit. 
- Then the next pass is to make sure that batch submits via the bulk submit. 

- [ ] Added the Jira Ticket to the title of the PR e.g. (DATAUP-69 Adds a PR template)

# Testing Instructions
- [ ] Changes available by spinning up a local test suite and doing X

# Dev Checklist:


lib/execution_engine2/sdk/job_submission_parameters.py
```diff
@@ -252,3 +252,6 @@ def __eq__(self, other):
 
     def __hash__(self):
         return hash(self._params())
+
+    def __repr__(self):
+        return str(self._params())
```
bio-boris, line 257: Was helpful for debugging tests

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -192,15 +193,19 @@ def _check_workspace_permissions(self, wsid):
...
-                f"User {self.sdkmr.user_id} doesn't have permission to run jobs in workspace {bad_ws}."
-            )
-            raise PermissionError(
-                f"User {self.sdkmr.user_id} doesn't have permission to run jobs in workspace {bad_ws}."
+        unique_not_none_not_zero_wsids = [wsid for wsid in set(wsids) if wsid]
+        if unique_not_none_not_zero_wsids:
+            perms = self.sdkmr.get_workspace_auth().can_write_list(
+                unique_not_none_not_zero_wsids
             )
+            bad_ws = [key for key in perms.keys() if perms[key] is False]
```
MrCreosote, line 201: Is there a reason for `is False` instead of `not perms[key]`?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +709,85 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+                runjob_param_wsid = runjob_param.get("wsid")
+                if runjob_param_wsid or runjob_param_wsid == 0:
+                    raise InvalidParameterForBatch()
+                # Do we do a deepcopy here in case the params point to the same obj?
+                runjob_params[i]["wsid"] = batch_wsid
+
+    def preflight(
+        self, runjob_params, batch_params=None, new_batch_job=False, as_admin=False
+    ):
+        if batch_params and not new_batch_job:
```
MrCreosote, line 760: Naively it seems to me that `batch_params` and `new_batch_job` are redundant - if `batch_params` is truthy you can just assume `new_batch_job` is True. I don't think I understand the purpose of the argument

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +709,85 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+        """
+        For batch jobs, check to make sure the job params do not provide a wsid other than None or 0
+        Then Modify the run job params to use the batch params wsid, which may be set to None
+        """
+        if batch_params is None:
+            batch_params = {}
+
+        if new_batch_job:
+            batch_wsid = batch_params.get("wsid")
+            for i, runjob_param in enumerate(runjob_params):
```
MrCreosote, line 750: Maybe I'm missing something here, but I don't think the enumerate is necessary - you could just do this, right?

```
for runjob_param in runjob_params:
    runjob_param["wsid"] = batch_wsid

File: lib/execution_engine2/sdk/EE2Runjob.py [hunk 5]
```diff
@@ -702,20 +709,85 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+                runjob_param_wsid = runjob_param.get("wsid")
+                if runjob_param_wsid or runjob_param_wsid == 0:
```
MrCreosote, line 752: why not just `if runjob_param_wsid is not None` ?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +709,85 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+            )
+        else:
+            raise IncorrectParamsException(
+                "runjob_params must be a mapping or a list of mappings"
+            )
+
+    @staticmethod
+    def _propagate_wsid_for_new_batch_jobs(runjob_params, batch_params, new_batch_job):
+        """
+        For batch jobs, check to make sure the job params do not provide a wsid other than None or 0
```
MrCreosote, line 742: Based on the code, it'll throw an exception if the wsid is 0

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Never mind - `batch-params` can still be `None` or `{}` for a batch job. Could we document that though? I'm guessing I'm not the only person that'll get confused

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +709,85 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+        Check a single job, a single batch job, or a retry_multiple request with a mix of different jobs.
+        """
+        if as_admin:
+            return self.sdkmr.check_as_admin(requested_perm=JobPermissions.WRITE)
+        # Batch params have the next highest
+        if new_batch_job:
+            if batch_params:
+                return self._check_workspace_permissions(batch_params.get("wsid"))
+            # No ws perms to check
+            return
```
MrCreosote, line 725: Due to the if block structure this return isn't necessary

MrCreosote (changes requested):
IIRC you're still working on tests so I haven't looked at those yet.

Code looks good, just some minor cleanup and docs.

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Which is probably the right thing to do since that's an illegal value and so indicates a bug in the calling code

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: The function returns "True" or "False" so it explicitly checks against that.  Seems a bit more readable too.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Ah good idea, this was causing issues for me when 0 was passed in.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Fixed comment

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Fixed

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Ah its  from when I was passing in the same params objects and they were getting modified and then throwing an exception on the next iteration of the loop. Fixed.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Added that to comments

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Fair enough

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: :+1:

MrCreosote:
code looks good; let me know when you're done with tests and I'll review those

bio-boris:
@MrCreosote I am done with the tests, I just added a bit more coverage for some cases, and realized I couldn't cover some of the cases yet.

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1324,8 +1338,9 @@ def test_run_job_batch(ee2_port, ws_controller, mongo_client):
             parent_job_id=parent_job_id,
             app_id=None,
             app_module=None,
+            wsid=job_batch_wsid,
         )
-        expected_sub_1["+KB_WSID"] = ""
+        expected_sub_1["+KB_WSID"] = f'"{job_batch_wsid}"'
```
MrCreosote, line 1343: Do we have a test case for a batch job without a workspace ID anymore?

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1478,9 +1499,10 @@ def test_run_job_batch_as_admin_with_job_reqs(ee2_port, ws_controller, mongo_cli
             "_id": ObjectId(parent_job_id),
             "user": USER_WRITE_ADMIN,
             "authstrat": "kbaseworkspace",
-            "wsid": 1,
+            "wsid": job_batch_wsid,
             "status": "created",
             "job_input": {
+                # "wsid": job_batch_wsid,
```
MrCreosote, line 1505: remove?

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1580,12 +1607,22 @@ def test_run_job_batch_fail_no_workspace_access_for_batch(ee2_port, ws_controlle
...
 
 
-def test_run_job_batch_fail_no_workspace_access_for_job(ee2_port):
+def test_run_job_batch_fail_no_allowed_wsid(ee2_port):
     params = [
         {"method": _MOD},
         {"method": _MOD, "wsid": 1},
     ]
     # this error could probably use some cleanup
+    err = InvalidParameterForBatch().__doc__
```
MrCreosote, line 1616: I would recommend using the actual error string here. If I fatfinger an unintended change to the error then this test won't catch it, but if I use the bare string it will.

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -615,12 +620,22 @@ class and its respective composed classes, and we don't reproduce all the error
 
     Tests both the run() and run_batch() methods.
     """
+    # These are extremely annoying to debug as they don't raise a stacktrace if a different exception type was thrown
+    # or let you know that it was an entirely different exception, or if the exception happened in the bulk version of the run
```
MrCreosote, line 624: In the java version of this code, it prints out the stacktrace if the exception type isn't the same - we could make the same update here. I just haven't needed it yet. You'd just add the stringified exception as the second argument to `assert`.

> or let you know that it was an entirely different exception

I don't understand this part - `assert type(got) == type(expected)` should tell you the two exception types are different

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -812,6 +831,7 @@ def _check_common_mock_calls_batch(mocks, reqs1, reqs2, parent_wsid, wsid):
         app=_APP_1,
         git_commit=_GIT_COMMIT_1,
         source_ws_objects=[_WS_REF_1, _WS_REF_2],
+        wsid=parent_wsid,
```
MrCreosote, line 834: I think we can remove the regular wsid argument from the method sig now, right?

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
```diff
@@ -242,10 +251,26 @@ def test_run_job(self, rq_mock, condor_mock):
...
+        condor_mock.run_job = MagicMock(return_value=si, side_effect=Exception("fail"))
+        runner.get_runjob()._finish_created_job = MagicMock(return_value=None)
+
+        with self.assertRaises(expected_exception=Exception):
+            runner.run_job(params=job)
+
+        # Condor Failure Case Coverage #2
+        with self.assertRaisesRegex(
+            expected_exception=RuntimeError,
+            expected_regex="Condor job not ran, and error not found. Something went wrong",
```
MrCreosote, line 269: s/ran/run/

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
```diff
@@ -515,31 +539,66 @@ def test_run_job_batch(self, rq_mock, condor_mock):
...
+        with self.assertRaises(InvalidParameterForBatch):
+            job_good = get_example_job_as_dict(
+                user=self.user_id, wsid=None, source_ws_objects=[]
+            )
+            job_bad = (
+                get_example_job(user=self.user_id, wsid=self.ws_id).to_mongo().to_dict()
+            )
+            job_bad["method"] = job["job_input"]["method"]
+            job_bad["app_id"] = job["job_input"]["app_id"]
+            job_bad["service_ver"] = job["job_input"]["service_ver"]
```
MrCreosote, line 580: Why don't you need to do this for `job_good` as well?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +710,95 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+
+        if new_batch_job:
+            batch_wsid = batch_params.get("wsid")
+            for runjob_param in runjob_params:
+                if runjob_param.get("wsid") is not None:
+                    raise InvalidParameterForBatch()
+                # Do we do a deepcopy here in case the params point to the same obj?
+                runjob_param["wsid"] = batch_wsid
+
+    def preflight(
```
MrCreosote, line 755: preflight should be a private method AFAICT, there's no reason to make it public is there?

MrCreosote (changes requested):

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -192,15 +193,20 @@ def _check_workspace_permissions(self, wsid):
...
     def _check_workspace_permissions_list(self, wsids):
-        perms = self.sdkmr.get_workspace_auth().can_write_list(wsids)
-        bad_ws = [key for key in perms.keys() if perms[key] is False]
-        if bad_ws:
-            self.logger.debug(
-                f"User {self.sdkmr.user_id} doesn't have permission to run jobs in workspace {bad_ws}."
-            )
-            raise PermissionError(
-                f"User {self.sdkmr.user_id} doesn't have permission to run jobs in workspace {bad_ws}."
+        # TODO Cover this in tests once you can execute multiple independent runs
```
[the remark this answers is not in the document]
bio-boris: Yes, when retrying multiple jobs they can all have different workspace ids.

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +710,95 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+        """
+        if batch_params is None:
+            batch_params = {}
+
+        if new_batch_job:
+            batch_wsid = batch_params.get("wsid")
+            for runjob_param in runjob_params:
+                if runjob_param.get("wsid") is not None:
+                    raise InvalidParameterForBatch()
+                # Do we do a deepcopy here in case the params point to the same obj?
```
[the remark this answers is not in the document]
bio-boris: The problem is that if its the same params object, then next iteration would throw an exception
params = {}
a = params
b = params
_propagate_wsid_for_new_batch_jobs([a,b])
After the first loop, both a and b have 'wsid' = the batch wsid, and the 2nd loop it would raise InvalidParameterForBatch() since B now "specifies" a wsid

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1247,23 +1251,32 @@ def test_run_job_batch(ee2_port, ws_controller, mongo_client):
...
         }
+        job_wsid = job1.get("wsid")
+        expected_job_wsid = expected_job1.get("wsid")
+
+        job_input_wsid = job1.get("wsid")
+        expected_job_input_wsid = expected_job1.get("wsid")
+
+        assert job_wsid == expected_job_wsid
+        assert job_input_wsid == expected_job_input_wsid
         assert job1 == expected_job1
```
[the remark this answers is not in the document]
bio-boris: This was used to greatly aid in debugging, as it was hard to tell which field didn't match which field from the thrown exception. Do you want to remove it?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +710,95 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+        # Single job runs
+        elif isinstance(runjob_params, dict):
+            return self._check_workspace_permissions(runjob_params.get("wsid"))
+        # Multiple independent job runs, think retry_multiple()
+        elif isinstance(runjob_params, list):
+            return self._check_workspace_permissions_list(
+                [job_param.get("wsid") for job_param in runjob_params]
+            )
+        else:
+            raise IncorrectParamsException(
```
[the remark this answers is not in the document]
bio-boris: Added `        if type(params) != dict:
            raise IncorrectParamsException("params must be a mapping")` 
to run()

So if the only way to get to this function is through 

run->preflight->_check_ws_perms
or
run_batch->Preflight->_check_ws_perms

And both of them check to see that the inputs are a dict or a list first, how do we test this exception by only testing the public api?

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: Hmm... when would that actually happen though? A user is typically working in a single workspace / narrative at a time.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: If code can't be reached via the public API then the code isn't needed and can be deleted

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: I see, ok. I assume that can't happen under normal usage though since the input is deserialized from JSON. It'd have to be someone calling against internals with that specific setup, so unlikely, right?

test/tests_for_integration/api_to_db_test.py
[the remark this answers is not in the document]
MrCreosote: I get something like this from the test output if there's a difference:
```
E             Full diff:
E               {
E                '_id': ObjectId('60e88ebefa84bddcb48de533'),
E                'authstrat': 'kbaseworkspace',
E                'batch_id': '60e88ebefa84bddcb48de532',
E                'batch_job': False,
E                'child_jobs': [],
E                'job_input': {'method': 'mod.meth',
E                              'narrative_cell_info': {'cell_id': 'cid',
E                                                      'run_id': 'rid',
E                                                      'tag': 'yourit',
E                                                      'token_id': 'tid'},
E                              'params': [{'foo': 'bar'}, 42],
E             -                'requirements': {'clientgroup': 'njs2',
E             ?                                                    -
E             +                'requirements': {'clientgroup': 'njs',
E                                               'cpu': 8,
E                                               'disk': 30,
E                                               'memory': 5},
E                              'service_ver': 'somehash',
E                              'source_ws_objects': ['1/1/1', '1/2/1']},
E                'retry_ids': [],
E                'retry_saved_toggle': False,
E                'scheduler_id': '123',
E                'scheduler_type': 'condor',
E                'status': 'queued',
E                'user': 'nouser',
E               }
```
Do you not get that?

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
bio-boris: When calling the api directly you can pass in any job ids you want for retry_multiple(). It seems wrong to not account for this behavior without this or preventing passing in multiple wsids to retry_multiple()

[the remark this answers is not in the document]
MrCreosote: I'm confused then - if you can submit multiple workspace IDs via `retry_multiple` why can't this function be tested?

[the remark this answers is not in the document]
bio-boris: Because right now they are calling run_job one at a time, and that one causes a seperate ws lookup for each job, instead of in 1 ws call, like this section will expect.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
bio-boris: Hmmmmmmm something about this just seems funny but alright, will just delete it then.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
bio-boris: It was happening during tests, which caused me to re-write some of the tests. It was already happening. In the future it's likely another test can make this mistake if they are not careful.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: `they` = `retry_multiple` I assume. Ok, that makes sense

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -702,20 +710,95 @@ def _get_run_job_params_from_existing_job(job: Job, user_id: str) -> Dict:
...
+    def _propagate_wsid_for_new_batch_jobs(runjob_params, batch_params, new_batch_job):
+        """
+        For batch jobs, check to make sure the job params do not provide a wsid other than None
+        Then Modify the run job params to use the batch params wsid, which may be set to None
+        """
+        if batch_params is None:
+            batch_params = {}
+
+        if new_batch_job:
+            batch_wsid = batch_params.get("wsid")
```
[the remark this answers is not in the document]
bio-boris: done

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: changed

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1187,11 +1189,12 @@ def test_run_job_batch(ee2_port, ws_controller, mongo_client):
         job2_params = {
             "method": "mod2.meth2",
             "app_id": "mod2/app2",
-            "wsid": 1,
+            # "wsid": 1,
```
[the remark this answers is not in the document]
bio-boris: done

test/tests_for_integration/api_to_db_test.py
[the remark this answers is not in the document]
bio-boris: No, I see

```
>           assert job1 == expected_job1
E           AssertionError: assert {'_id': ObjectId('60e8ac217ff839f17c75521c'),\n 'authstrat': 'kbaseworkspace',\n 'batch_id': '60e8ac217ff839f17c75521b',\n 'batch_job': False,\n 'child_jobs': [],\n 'job_input': {'method': 'mod.meth',\n               'narrative_cell_info': {'cell_id': 'cid',\n                                       'run_id': 'rid',\n                                       'tag': 'yourit',\n                                       'token_id': 'tid'},\n               'params': [{'foo': 'bar'}, 42],\n               'requirements': {'clientgroup': 'njs',\n                                'cpu': 8,\n                                'disk': 30,\n                                'memory': 5},\n               'service_ver': 'somehash',\n               'source_ws_objects': ['1/1/1', '1/2/1'],\n               'wsid': 2},\n 'retry_ids': [],\n 'retry_saved_toggle': False,\n 'scheduler_id': '123',\n 'scheduler_type': 'condor',\n 'status': 'queued',\n 'user': 'nouser',\n 'wsid': 2} == {'_id': ObjectId('60e8ac217ff839f17c75521c'),\n 'authstrat': 'kbaseworkspace',\n 'batch_id': '60e8ac217ff839f17c75521b',\n 'batch_job': False,\n 'child_jobs': [],\n 'job_input': {'method': 'mod.meth',\n               'narrative_cell_info': {'cell_id': 'cid',\n                                       'run_id': 'rid',\n                                       'tag': 'yourit',\n                                       'token_id': 'tid'},\n               'params': [{'foo': 'bar'}, 42],\n               'requirements': {'clientgroup': 'njs',\n                                'cpu': 8,\n                                'disk': 30,\n                                'memory': 5},\n               'service_ver': 'somehash',\n               'source_ws_objects': ['1/1/1', '1/2/1'],\n               'wsid': 2},\n 'retry_ids': [],\n 'retry_saved_toggle': False,\n 'scheduler_id': '123',\n 'scheduler_type': 'condor',\n 'status': 'queued',\n 'user': 'nouser',\n 'wsid': 'cat'}
```
Unless I run from cmd line, then I see

```
E                'retry_saved_toggle': False,
E                'scheduler_id': '123',
E                'scheduler_type': 'condor',
E                'status': 'queued',
E                'user': 'nouser',
E             -  'wsid': 'cat',
E             ?          ^^^^^
E             +  'wsid': 2,
E             ?          ^
E               }
```
Sometimes there is a diff available and that shows the difference in a GUI viewer

Do you want me to delete it?

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1292,9 +1305,10 @@ def test_run_job_batch(ee2_port, ws_controller, mongo_client):
             "_id": ObjectId(parent_job_id),
             "user": USER_NO_ADMIN,
             "authstrat": "kbaseworkspace",
-            "wsid": 2,
+            "wsid": job_batch_wsid,
             "status": "created",
             "job_input": {
+                # "wsid": job_batch_wsid,
```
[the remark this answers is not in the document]
bio-boris: done

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: In which case they'll get a failure, I assume, so that seems ok

test/tests_for_integration/api_to_db_test.py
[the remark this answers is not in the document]
MrCreosote: That's poopy, I guess pycharm doesn't integrate pytest well?

Up to you, but I'd at least note why the lines are there and that they're redundant to the 'master' assert, and I'd also delete or fix one of the two sets of asserts, because they're asserting on the same thing. I think the 2nd one was supposed to assert on the job input wsid but it's not - it's still doing the job wsid.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: It could be seen as future proofing the code for new uses of the method. If that's the intent I would drop the final else and just end the method with
```
assert 0, "called method with invalid params"
```
 or something like that

test/tests_for_integration/api_to_db_test.py
bio-boris: I don't think we ever had a test case for that, as job_batch_params always specifies a wsid in each of these tests, so it should always have a workspace id. I can write a test, but currently having an issue with adding another jobrun

no_wsid_ret = ee2.run_job_batch([job1_params, job2_params], job_batch_params_no_wsid)

results in

```
>                   raise ServerError(**err["error"])
E                   installed_clients.baseclient.ServerError: Server error: -32000. ()
E                   Traceback (most recent call last):
E                     File "/ee2/lib/execution_engine2/execution_engine2Server.py", line 107, in _call_method
E                       result = method(ctx, *params)
E                     File "/ee2/lib/execution_engine2/execution_engine2Impl.py", line 367, in run_job_batch
E                       job_ids = mr.run_job_batch(
E                     File "/ee2/lib/execution_engine2/sdk/SDKMethodRunner.py", line 328, in run_job_batch
E                       return self.get_runjob().run_batch(
E                     File "/ee2/lib/execution_engine2/sdk/EE2Runjob.py", line 381, in run_batch
E                       self._add_job_requirements(params, bool(as_admin))  # as_admin checked above
E                     File "/ee2/lib/execution_engine2/sdk/EE2Runjob.py", line 421, in _add_job_requirements
E                       job[_JOB_REQUIREMENTS] = jrr.resolve_requirements(
E                     File "/ee2/lib/execution_engine2/utils/job_requirements_resolver.py", line 412, in resolve_requirements
E                       cat_reqs_all = self._get_catalog_reqs(module_name, function_name, catalog_cache)
E                     File "/ee2/lib/execution_engine2/utils/job_requirements_resolver.py", line 469, in _get_catalog_reqs
E                       group_config = catalog_cache.lookup_job_resource_requirements(
E                     File "/ee2/lib/execution_engine2/utils/catalog_cache.py", line 91, in lookup_job_resource_requirements
E                       ] = self.get_catalog().list_client_group_configs(
E                     File "<string>", line 3, in list_client_group_configs
E                     File "/miniconda-latest/lib/python3.8/unittest/mock.py", line 1081, in __call__
E                       return self._mock_call(*args, **kwargs)
E                     File "/miniconda-latest/lib/python3.8/unittest/mock.py", line 1085, in _mock_call
E                       return self._execute_mock_call(*args, **kwargs)
E                     File "/miniconda-latest/lib/python3.8/unittest/mock.py", line 1142, in _execute_mock_call
E                       result = next(effect)
E                     File "/miniconda-latest/lib/python3.8/unittest/mock.py", line 388, in __next__
E                       return next(self.obj)
E                   StopIteration
```
Which I am trying to figure out... Does it have something to do with the mock side effects not returning enough values?

test/tests_for_integration/api_to_db_test.py
bio-boris: done

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
bio-boris: Is that better than throwing an exception? And the code is still not covered in that case as well, right?

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: From _create_job ? I don't think so, it's required to build up a job to compare to.

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
bio-boris: Removed from job_bad

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
bio-boris: ok

test/tests_for_integration/api_to_db_test.py
MrCreosote: > I don't think we ever had a test case for that

Looks like you're right

MrCreosote: Are you doing that in the same test? If so, and the test mocks have side effects, then yeah, that's what's happening. I (almost) always only run 1 method per test, and factor out any common setup / result or mock checking code into helper methods.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: yeah, because you can have the interpreter remove asserts.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: No, from `_check_common_mock_calls_batch` since all the wsids should be the parent wsid

test/tests_for_integration/api_to_db_test.py
bio-boris: Yes. Is it possible to factor out any common setup with these mocks? Is it possible to move these patching set code into another function? Or should I keep adding things to the side_effects. Or do you want me to copy and paste this test and then just change the part with the wsid?

test/tests_for_integration/api_to_db_test.py
bio-boris: ok

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: Here's an example


```
tests_for_integration/api_to_db_test.py:1608 (test_run_job_batch_fail_no_allowed_wsid)
('Workspace ids are not allowed in RunJobParams in Batch Mode',) != (InvalidParameterForBatch('Workspace ids are not allowed in RunJobParams in Batch Mode'),)

Expected :(InvalidParameterForBatch('Workspace ids are not allowed in RunJobParams in Batch Mode'),)
Actual   :('Workspace ids are not allowed in RunJobParams in Batch Mode',)
<Click to see difference>

ee2_port = '49193'

    def test_run_job_batch_fail_no_allowed_wsid(ee2_port):
        params = [
            {"method": _MOD},
            {"method": _MOD, "wsid": 1},
        ]
        # this error could probably use some cleanup
        err = InvalidParameterForBatch("Workspace ids are not allowed in RunJobParams in Batch Mode")
>       _run_job_batch_fail(ee2_port, TOKEN_NO_ADMIN, params, {"wsid": 1}, err)

api_to_db_test.py:1616: 
```
When I called the function wrong

```
    err = InvalidParameterForBatch("Workspace ids are not allowed in RunJobParams in Batch Mode")
    _run_job_batch_fail(ee2_port, TOKEN_NO_ADMIN, params, {"wsid": 1}, err)
```
instead of
```
    err = "Workspace ids are not allowed in RunJobParams in Batch Mode"
    _run_job_batch_fail(ee2_port, TOKEN_NO_ADMIN, params, {"wsid": 1}, err)
```

test/tests_for_integration/api_to_db_test.py
MrCreosote: > Is it possible to factor out any common setup with these mocks? Is it possible to move these patching set code into another function? 

Should be, I did it with the non-batch functions (and I think some of the batch functions as well. 

> Or should I keep adding things to the side_effects.

I personally avoid that like the plague - the tests get too hard to understand if you're splitting the side effects between multiple calls.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: That seems pretty straightforward - it's telling you it returned a string rather than an exception
```

Expected :(InvalidParameterForBatch('Workspace ids are not allowed in RunJobParams in Batch Mode'),)
Actual   :('Workspace ids are not allowed in RunJobParams in Batch Mode',)
```

test/tests_for_integration/api_to_db_test.py
MrCreosote: Although another approach might be to C&P and then factor out common stuff in another PR

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: image

image

I was able to reproduce what I was talking about here

MrCreosote: E.g. you don't know which part of the test is throwing the exception. The fix for that would be

> In the java version of this code, it prints out the stacktrace if the exception type isn't the same - we could make the same update here. I just haven't needed it yet. You'd just add the stringified exception as the second argument to assert.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: Hmm, now I'm reviewing this again this definitely seems like a stumbling block for future coders - if they submit an uncovered scenario it'll just drop through the method and seem like the perms were checked when they weren't. I'd at least put the assert in to protect against that.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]

test/tests_for_integration/api_to_db_test.py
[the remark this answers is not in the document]
MrCreosote: :+1:

test/tests_for_integration/api_to_db_test.py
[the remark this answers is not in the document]
MrCreosote: :+1:

test/tests_for_integration/api_to_db_test.py
MrCreosote: :+1:

test/tests_for_integration/api_to_db_test.py
MrCreosote: :+1:

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: :+1:

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
MrCreosote: :+1:

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
MrCreosote: :+1:

lib/execution_engine2/sdk/EE2Runjob.py

MrCreosote:
I think there's some comments in the prior review that could use another look:
https://github.com/kbase/execution_engine2/pull/402#discussion_r669975782
https://github.com/kbase/execution_engine2/pull/402#discussion_r667195856

MrCreosote:
Looks good otherwise.

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
bio-boris: I've never used assert outside of tests because they can get removed during compilation.  So again I don't understand why you want to assert here instead of throwing an exception

test/tests_for_integration/api_to_db_test.py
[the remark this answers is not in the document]
bio-boris: deleted

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: So if you want to use an exception that's fine, it's not worth the book we've written at this point :P

That being said, this is exactly what asserts are for - when you're debugging / testing code, you leave the asserts in to catch mistakes. When you're ready for deploy, you can compile them out.

MrCreosote (approved):
LGTM

MrCreosote:
Wait tests are failing

MrCreosote:
Oh and the dev checklist isn't filled out

PR #406, opened 2021-07-16
DATAUP-530 Refactor to bulk insert

bio-boris:

* In favor of https://github.com/kbase/execution_engine2/pull/403
* Use a mongoengine bulk create (with no limit on number of docs that can be written)
* Use a pymongo bulk update job status
* Change kafka creation message order
* Batch submits via the bulk submit.

# Testing Instructions
* Deployed on CI

# Dev Checklist:


lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +263,27 @@ def check_if_already_finished(job_status):
             return True
         return False
 
+    def update_jobs_to_queued(self, job_ids, scheduler_ids):
```
bio-boris, line 266: Currently unused

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -371,17 +493,30 @@ def run_batch(
...
 
+        preflight_begin = time.time()
         self._preflight(
             runjob_params=params,
             batch_params=batch_params,
             new_batch_job=True,
             as_admin=as_admin,
         )
-
+        print("Preflight took", time.time() - preflight_begin)
```
bio-boris, line 503: Leaving timings in for the next few prs

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +263,33 @@ def check_if_already_finished(job_status):
...
```
MrCreosote, line 266: Since there's a 1:1 mapping of job id to scheduler id maybe a dict or list of tuples would make more sense here? Then there's no way that 2 different size lists could be submitted

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +263,33 @@ def check_if_already_finished(job_status):
...
+        for i, job_id in enumerate(job_ids):
+            bulk_operations.append(
+                UpdateOne(
+                    {"_id": ObjectId(job_id)},
+                    {
+                        "$set": {
+                            "status": Status.queued.value,
+                            "queued": now,
+                            "scheduler_id": scheduler_ids[i],
+                            "scheduler_type": "condor",
```
MrCreosote, line 278: This is coupling MongoUtils to Condor - it might make more sense to pass this in, either in the constructor or in the method

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -420,6 +447,26 @@ def update_job_status(self, job_id, status, msg=None, error_message=None):
...
+    def insert_jobs(self, jobs_to_insert: List[Job]) -> List[ObjectId]:
+        """
+        Insert multiple job records using MongoEngine
+        :param jobs_to_insert: Multiple jobs to insert at once
+        :return: List of job ids from the insertion
+        """
+        # TODO Look at pymongo write_concerns that may be useful
+        # TODO see if pymongo is faster
+
+        inserted = Job.objects.insert(doc_or_docs=jobs_to_insert, load_bulk=False)
```
MrCreosote, line 459: Can you remind me how insert failures work? Can you get partial successes? How would partial failures be handled in the calling method?

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -420,6 +447,26 @@ def update_job_status(self, job_id, status, msg=None, error_message=None):
...
+
+        return inserted
+        # TODO: Send Kafka messages here or elsewhere
```
MrCreosote, line 462: I'd recommend not coupling mongoutils and kafka, personally

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +248,150 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+        try:
+            submission_ids = self._submit_multiple(job_submission_params)
+            self.logger.debug(f"condor_time submit = {time.time() - condor_time}")
+            return submission_ids
+        except Exception as e:
+            self._abort_multiple_jobs(job_ids)
+            raise e
+
+    def _update_to_queued_multiple(self, job_ids, scheduler_ids):
+        # TODO Unused
```
MrCreosote, line 318: I'm not sure what this comment is referring to?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +248,150 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+        if len(job_ids) != len(scheduler_ids):
+            raise Exception(
+                "Need to provide the same amount of job ids and scheduler_ids"
+            )
+
+        # TODO RETRY FOR RACE CONDITION OF RUN/CANCEL
+        # TODO PASS QUEUE TIME IN FROM SCHEDULER ITSELF?
+        # TODO PASS IN SCHEDULER TYPE?
```
MrCreosote, line 326: yeah, I would do this

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +248,150 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+
+    def _update_to_queued_multiple(self, job_ids, scheduler_ids):
+        # TODO Unused
+        if len(job_ids) != len(scheduler_ids):
+            raise Exception(
+                "Need to provide the same amount of job ids and scheduler_ids"
+            )
+
+        # TODO RETRY FOR RACE CONDITION OF RUN/CANCEL
+        # TODO PASS QUEUE TIME IN FROM SCHEDULER ITSELF?
```
MrCreosote, line 325: If the time in mongo can match the time from Condor that seems better, but it's probably only a few seconds, so meh? Could wait for a later PR for sure, maybe just make an issue?

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +248,150 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+            raise e
+
+    def _update_to_queued_multiple(self, job_ids, scheduler_ids):
+        # TODO Unused
+        if len(job_ids) != len(scheduler_ids):
+            raise Exception(
+                "Need to provide the same amount of job ids and scheduler_ids"
+            )
+
+        # TODO RETRY FOR RACE CONDITION OF RUN/CANCEL
```
MrCreosote, line 324: Can you explain this? This sounds like it could cause problems

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -326,28 +452,24 @@ def _create_batch_job(self, wsid, meta):
...
-                )
-                self._abort_child_jobs(child_jobs)
-                raise e
+        run_multiple = time.time()
+        child_jobs = self._run_multiple(params)
+        self.logger.debug(f"run_multiple  = {time.time() - run_multiple}")
 
+        batch_save = time.time()
         batch_job.child_jobs = child_jobs
         self.sdkmr.save_job(batch_job)
```
MrCreosote, line 471: Should this be updated to just `$push` the child jobs to avoid race conditions? I'm not sure

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +248,150 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+        Creates a job record and creates the job submission params
+        """
+
+        job_id = self._init_job_rec(self.sdkmr.get_user_id(), params)
+        self.logger.debug(
+            f"User {self.sdkmr.get_user_id()} attempting to run job {params[_METHOD]} {params}"
+        )
+        return self._generate_job_submission_params(job_id, params)
+
+    def _run_multiple(self, runjob_params):
```
MrCreosote, line 262: So now there are two job running paths: `_run` and `_run_multiple`, and `_run_multiple` can do everything that `_run` can do. It seems to me we could simplify the code by getting rid of `_run` and have everything use the `_run_multiple` code path.

MrCreosote (changes requested):
Mostly looks pretty reasonable. I have a few questions re error handling and some internal api suggestions

MrCreosote:
Oh, there's a couple of checklist things that aren't checked or marked as N/A, missed that

lib/execution_engine2/db/MongoUtil.py
bio-boris: Seems like a small amount of code to change if and when we decide to change schedulers, doesn't seem worth it right now to pass it in.

lib/execution_engine2/db/MongoUtil.py
bio-boris: Before mongo 4.2 there aren't transactions, so yes there are partial successes and failures. If there is a failure, there should be some records inserted and some nothing, and then the method returns. A "created" jobs reaper could reap jobs that are in created state for too long.

lib/execution_engine2/db/MongoUtil.py
bio-boris: One advantage is that the user doesn't have to remember to create a kafka message, but yea i didn't end up doing that in this PR

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: I don't think it matters yea

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: On second thought I don't think there can be a race condition, because you don't know the job ids to cancel until the method returns.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: In theory you don't know the job id until the method returns, so you wouldn't be able to do it unless you were monitoring the db for the new records. But yea we could do that.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Right, but then we lose the more detailed error message handling for single run jobs

lib/execution_engine2/db/MongoUtil.py
bio-boris: Alright, we can do that, and then replace the check with a check to make sure that one of the job id and scheduler id combos doesn't have a None in it?

MrCreosote: Sounds good to me

lib/execution_engine2/db/MongoUtil.py
MrCreosote: [Reducing code coupling is a good thing in general](https://en.wikipedia.org/wiki/Coupling_(computer_programming)), and I mean, this is a 5 min fix so I'd just do it

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Say I submit 10 jobs and one fails. How does the code handle that, and what would the client programmer see?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: But the user may not want to sent a kafka message

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: But if you thread everything after the initial mongo insert then there could be, right?

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Might as well, easy fix

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Ok, how about making a ticket to look into it further?

I would be interested in knowing how much we're degrading the error messaging by switching `run_job_batch` to use `_run_multiple` vs `_run` though

lib/execution_engine2/db/MongoUtil.py
bio-boris: I'd expect a mongoengine error text to be shown in the ServerError

lib/execution_engine2/db/MongoUtil.py
bio-boris: We removed the entire scheduler class to inherit from, so I am assuming we aren't going to have any new schedulers anytime soon, seems like YAGNI to me.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Right.. So if while you are running these jobs, you hit cancel, then there's a thread that's still going and updating the jobs to "queued".. So cancel won't actually cancel them all, and it might not be able to kill the jobs from condor because it hasn't been submitted to condor yet.

lib/execution_engine2/db/MongoUtil.py
bio-boris: done

lib/execution_engine2/db/MongoUtil.py
bio-boris: Well the user would be wrong not to do so , because we are supposed to send kafka messages on status transitions.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: removed

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: done

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Right, so even if we thread the updating code as it stands now, there is still a small window in which it's possible to call update on a job that should be cancelled. It would probably be an even smaller window than this pr.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: done

bio-boris: Wasn't really an easy fix, there was a mongoengine connection not defined error, so had to move this to an sdkmr method. Then this screwed up some of the tests
# Not sure how to get the final parent job anymore to do this test HALP

lib/execution_engine2/db/MongoUtil.py
MrCreosote: If you make a class with `run_job`, `get_job_resource_info`, and `cancel_job` methods you've just made a new scheduler class that you can plug right in with ducktyping, so I don't think removing the abstract class really affects things. It'd be easy to make a new one based on the above methods if we wanted to anyway.

YAGNI is for stuff like features and optimization, not code quality issues like cohesion and separation of concerns. You always want to do that when possible.

I'm honestly a little confused why this this so controversial - we've spent more time talking about it than it would have taken to just make the change at this point. Given the change is so simple, what's your concern about making it?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Ok, so basically the 9 created jobs are abandoned and some sort of reaper would have to terminate them later I guess?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: For this particular usage of the module, not necessarily all possible usages of the module. This gets back into separation of concerns and making a module responsible for one thing - in this case interfacing with mongo.

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Oh, according to https://github.com/kbase/execution_engine2/pull/406#discussion_r672770748 you did make the change. I'll take a look at the diff in a sec

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Hmm, so for a lot of jobs where the thread is running for a while, a user could cancel a job before the thread has set the update state, and then the thread would come along and blow away the canceled state.
Are there any other fields other than the state that could be wrong? Like a finished timestamp?

Right now I'm thinking that this is out of scope for this PR and we should just make an issue and deal with it in a later PR

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: If the call to `add_child_jobs` is just `(job_id, list_job_child_job_ids)` (haven't looked at the diff yet) you don't need the parent job any more - you can just do something like
```
sdkmr.add_child_jobs.assert_called_once_with(job_id, list_child_job_ids)
```

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Did you want to make an issue for this or just leave it as is? I don't have a strong opinion

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: I think this question is still open yeah?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
MrCreosote: :+1:

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +264,47 @@ def check_if_already_finished(job_status):
...
+                    },
+                )
+            )
+        mongo_collection = self.config["mongo-jobs-collection"]
+        if bulk_operations:
+            with self.pymongo_client(mongo_collection) as pymongo_client:
+                bwr = pymongo_client[self.mongo_database][mongo_collection].bulk_write(
+                    bulk_operations, ordered=False
+                )
+                assert bwr.modified_count == len(job_id_pairs)
```
MrCreosote, line 304: Is this guaranteed to be true if `bulk_write` returns without throwing an error?

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +264,47 @@ def check_if_already_finished(job_status):
...
+        now = time.time()
+        for job_id_pair in job_id_pairs:
+            if None in job_id_pairs:
+                raise InvalidStatusTransitionException
+            bulk_operations.append(
+                UpdateOne(
+                    {
+                        "_id": ObjectId(job_id_pair.job_id),
+                        "status": {
+                            "$in": [Status.created.value, Status.estimating.value]
```
MrCreosote, line 285: I'm wondering about the estimating state here. I'm not sure how a job gets to or leaves that state so it's not clear to me whether allowing a job that still in the estimating phase to move to the queued phase makes sense.

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Ok, I see it's passing in the parent job. In that case, you'd have to get the parent job from the prior save call where it's passed into `SDKMR.save_job()`

MrCreosote: Ok, I see it's passing in the parent job. In that case, you'd just manually construct your own job object and check that it matches the job object that's passed in.

MrCreosote: To get that job you do something like
```
final_got_parent_job = sdkmr.add_child_jobs.call_args_list[0][0][0]
```

MrCreosote (changes requested):
Couple questions on the new stuff

MrCreosote (changes requested):
couple questions on the new stuff

MrCreosote:
Well that's weird, not quite sure what happened there

lib/execution_engine2/db/MongoUtil.py
bio-boris: The lifecycle is Create->Estimating->Queued . Estimated is actually queued in condor, but its a second job.. We could remove it for now. In the future,  the function can be renamed to "update_jobs_to_queued_or_estimating" and allow for updating to estimating here, or we can create a separate function instead.

lib/execution_engine2/db/MongoUtil.py
bio-boris: it is NOT guaranteed true, because if one of the UpdateNow statements doesn't work, e.g. the job is no longer in a state able to be updated to "queued", it should throw an exception and cause these jobs to get cancelled

lib/execution_engine2/db/MongoUtil.py
bio-boris: Right,  this is covered by the reaper,  unless we can figure out how to terminate the jobs right here and then, or rollback them somehow.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Created a ticket https://kbase-jira.atlassian.net/browse/DATAUP-537

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: It shouldn't blow away the canceled state, since the Update statement can't work, since it updates only records that are in the state of "Created", and if its in the state of cancelled, it can't update.

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Well, the code could check if the inserted count equals the incoming job count and 

* retry any failed jobs for some number of retries
* remove all the non-failed jobs from the db and throw an error
* terminate all the non-failed jobs

Since this is job creation we're still in synchronous mode so I think all of those are safe options at least from the standpoint of feedback to the user. Of course if the DB is down or something and stays down there's not much that can be done

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Right, I wrote that before I looked at the most recent diffs

MrCreosote: So is this issue fixed? It seems like if a race condition does happen and a job is cancelled before it's updated the mongo record stays as cancelled

MrCreosote: Oh but the condor job might keep running

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Ok, but for the purposes of this PR i'd like to know how the error messaging is going to change for batch jobs, since we're actually making the change right now

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Sorry, I'm confused - it seems like from what you're saying there are 2 scenarios:

* The job returns without an error and the assert is true
* The job throws and exception and the assert statement is skipped

From that is seems to me that the assert has to be true

lib/execution_engine2/db/MongoUtil.py
MrCreosote: My preference would be to remove it until we actually implement estimation

lib/execution_engine2/db/MongoUtil.py
bio-boris: Seems like a lot of work for this PR, maybe this should go into another PR?

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: * If I am understanding the way mongo updateone statements believe, I believe so.
*  It's not possible to cancel the condor job if we don't know its scheduler_id.. It will silently fail and send some poorly formed kafka messages
* The condor job has the job cancellation checker though, so it should get cancelled that way if it checks the jobs cancellation status in ee2 db

lib/execution_engine2/db/MongoUtil.py
MrCreosote: I'm not actually advocating for any particular solution right now. Let me ask this: what would happen in a partial failure situation assuming we were running this code?

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Ok, that last one makes this seem fine. I didn't notice a test showing that jobs not in the created state don't get updated though, maybe add that?

lib/execution_engine2/db/MongoUtil.py
bio-boris: There are 3 scenarios
* The bulk write returns without error, and the assert is true
* The bulk write returns without error, and the number of Updated documents match is less than the number of jobs going in (one of the documents is now in a cancelled state, so it can't update it the document, so it doesn't count towards the updateCount)
* The job throws an exception and the assert statement is skipped

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: * If jobs fail the update to queued, instead of a mongoengine error, it will be a pymongo error. Kafka messages will be sent only after all updates are done, not after each update. 
* If jobs fail after add_child_jobs, it throws a slightly different mongoengine error instead
* Currently the entire job lifecycle is done before a submission fails, So create, submit, update to queued, then eventually a submission failure, followed by terminating all jobs. The new way to fail is to fail in between each of these lifecycle stages. For batch, I don't think it makes much of a difference right now.

After looking more closely, I think the error messaging will only significantly change when condor is unavailable, and when job submission is in a new thread, and the cell will transition to terminated state after clicking "Run" or "Retry" and then the cell will have to wait for the narrative to update the next job status.

lib/execution_engine2/db/MongoUtil.py
bio-boris: done

lib/execution_engine2/db/MongoUtil.py
bio-boris: It calls collection.insert_many
```
        with set_write_concern(self._collection, write_concern) as collection:
            insert_func = collection.insert_many
            if return_one:
                raw = raw[0]
                insert_func = collection.insert_one

        try:
            inserted_result = insert_func(raw)
            ids = [inserted_result.inserted_id] if return_one else inserted_result.inserted_ids
        except pymongo.errors.DuplicateKeyError as err:
            message = 'Could not save document (%s)'
            raise NotUniqueError(message % six.text_type(err))
        except pymongo.errors.BulkWriteError as err:
            # inserting documents that already have an _id field will
            # give huge performance debt or raise
            message = 'Document must not have _id value before bulk write (%s)'
            raise NotUniqueError(message % six.text_type(err))
        except pymongo.errors.OperationFailure as err:
            message = 'Could not save document (%s)'
            if re.match('^E1100[01] duplicate key', six.text_type(err)):
                # E11000 - duplicate key error index
                # E11001 - duplicate key on update
                message = 'Tried to save duplicate unique keys (%s)'
                raise NotUniqueError(message % six.text_type(err))
            raise OperationError(message % six.text_type(err))
```
And fails, which can leave the db in a state with some jobs inserted, and some not inserted.  As for what happens when insert_many is called is

```
        blk = _Bulk(self, ordered, bypass_document_validation)
        blk.ops = [doc for doc in gen()]
        blk.execute(write_concern, session=session)
        return InsertManyResult(inserted_ids, write_concern.acknowledged)
```

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Changed the assert and added tests.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: I think to just leave it

lib/execution_engine2/db/MongoUtil.py
MrCreosote: I mean what would happen to the jobs, and what would happen for the user

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Fair enough

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Ok, so am I right in thinking that the error messaging changes are basically just from an error that your average user isn't going to be able to understand / do anything about to a different error with the same properties?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Ok, for the second one then that assert is going to get triggered, but that's an expected scenario right? There's nothing wrong with the mod count being < `job_id_pairs`, right?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
MrCreosote: It's an `if` not an `assert` now

lib/execution_engine2/db/MongoUtil.py
bio-boris: The user gets a failure message during the submit, and some of the jobs are saved partially in a created state, and stay in that state until a reaper gets to them.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Probably. 
My guess would be though in the scenario that if condor is down, and the cell crashes right away, the user will try a few times and give up.. But if the cell crashes later, they might keep submitting the jobs since they see the cell transition to a queued state for a bit.

lib/execution_engine2/db/MongoUtil.py
bio-boris: It's wrong in that
* The wrong number of kafka messages would get sent out for state transitions to Updated, but yea, the damage is done, the jobs are already sent to HTCondor. This will thrown an exception, and we could in theory use this exception  to call a `condor_rm` on these jobs, so they don't take up the user's slots.

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -57,6 +58,39 @@ def test_init_ok(self):
         mongo_util = self.getMongoUtil()
         self.assertTrue(set(class_attri) <= set(mongo_util.__dict__.keys()))
 
+    def test_update_jobs_success(self):
```
MrCreosote, line 61: I don't quite undestand this test - it just saves a couple jobs and that's it?

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -57,6 +58,39 @@ def test_init_ok(self):
...
+        job = get_example_job(status=Status.created.value)
+        job2 = get_example_job(status=Status.created.value)
+        job_id1 = job.save().id
+        job_id2 = job2.save().id
+        job_ids = [job_id1, job_id2]
+        scheduler_ids = [job.scheduler_id, job2.scheduler_id]
+        jobs_to_update = list(map(JobIdPair, job_ids, scheduler_ids))
+        print(jobs_to_update)
+        assert Status.created.value in [job.status, job2.status]
```
MrCreosote, line 70: I don't think this is doing what you want - if one of the jobs has a created status but the other doesn't this will pass

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -57,6 +58,39 @@ def test_init_ok(self):
...
+        job3 = get_example_job(status=Status.terminated.value)
+
+        job_id1 = job.save().id
+        job_id2 = job2.save().id
+        job_id3 = job3.save().id
+        job_ids = [job_id1, job_id2, job_id3]
+        scheduler_ids = [job.scheduler_id, job2.scheduler_id, job3.scheduler_id]
+        jobs_to_update = list(map(JobIdPair, job_ids, scheduler_ids))
+        print(jobs_to_update)
+        assert Status.created.value in [job.status, job2.status, job3.status]
```
MrCreosote, line 86: Same as above re the test not quite working

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -923,12 +926,16 @@ def _check_common_mock_calls_batch(mocks, reqs1, reqs2, parent_wsid):
...
-    final_got_parent_job = sdkmr.save_job.call_args_list[4][0][0]
-    assert_jobs_equal(final_got_parent_job, final_expected_parent_job)
+    # Test to see if add_child jobs is called with correct batch_container and children
+    expected_batch_container = Job()
+    expected_batch_container.id = ObjectId(_JOB_ID)
+    expected_batch_container.user = _USER
+
+    batch_job = sdkmr.add_child_jobs.call_args_list[0][1]["batch_job"]
+    sdkmr.add_child_jobs.assert_called_once_with(
+        batch_job=expected_batch_container, child_jobs=[_JOB_ID_1, _JOB_ID_2]
```
MrCreosote, line 936: This relies on equality, which doesn't work for `Job` objects. You'll have to pull the `call_args_list` apart and test the parts individually, unfortunately.

MrCreosote (changes requested):
Taking a look at the newer changes

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Buuut tests are passing... huh that's weird

MrCreosote: Ok, I don't see how this is passing, not only from the equality issue, but also because the batch jobs have a bunch more fields than just `id` and `user`

MrCreosote: Oh, I bet it's because they have the same object ID, which is what `Job` equality is based on. So this test doesn't actually check that the call is corrent, but the `assert_jobs_equal` line below does. So this is ok as is.

MrCreosote: Actually if you could add a comment here explaining that the assert below is necessary because of how equality works for `Job` objects and that the assert_called_once_with doesn't correctly check the job object that'd help the next person that comes along

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Does the reaper exist now, or is that a future PR? How often does the reaper run?

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: Ok. IMO then we should move stuff that uses _run to _run_multiple, but that doesn't need to happen in this PR.

lib/execution_engine2/db/MongoUtil.py
MrCreosote: I mean they should cancel themselves soon enough anyway right

MrCreosote: So would the right thing to do be just return the IDs of the jobs that updated correctly from MongoUtils and just sent the Kafka messages for those?

test/tests_for_db/ee2_MongoUtil_test.py
bio-boris: Looks like either I forgot to write this test or deleleted it?? Hmm. Updated it

test/tests_for_db/ee2_MongoUtil_test.py
bio-boris: Yea I don't know what I was thinking here, these tests are pretty bad right now, will fix em.

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: Added

lib/execution_engine2/db/MongoUtil.py
bio-boris: It [exists](https://github.com/kbase/execution_engine2/pull/412/files), just is not deployed and set up to run every X mins.

lib/execution_engine2/db/MongoUtil.py
bio-boris: Yea, we could query mongo again and then send the kafka messages, that would be possible. Uh in theory they will cancel themselves, but then somewhat affects your priority in the queue and your overall score for jobs.

test/tests_for_db/ee2_MongoUtil_test.py
bio-boris: fixed

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Ok, so you're working on it

MrCreosote: All right, do you think leaving things as is is reasonable assuming the reaper gets done?

lib/execution_engine2/db/MongoUtil.py
MrCreosote: But that happens for everyone that cancels jobs, right? The only difference here is that the person cancelled it quickly.

test/tests_for_sdkmr/EE2Runjob_test.py
MrCreosote: Looks like there's some sort of interpolation in the comment?
```
Job objects
    (  print(jobs_to_update) because 
```

test/tests_for_db/ee2_MongoUtil_test.py
MrCreosote: :+1:

test/tests_for_db/ee2_MongoUtil_test.py
MrCreosote: :+1:

test/tests_for_db/ee2_MongoUtil_test.py
MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
bio-boris: yes

lib/execution_engine2/db/MongoUtil.py
bio-boris: No, for the person that cancelled it quickly the cancel jobs endpoint cannot send the cancellation signal to condor, since the cancel_jobs_endpoint doesn't know the condor job id yet until after this update statement is complete.

test/tests_for_sdkmr/EE2Runjob_test.py
bio-boris: fixed

MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
MrCreosote: :+1:

lib/execution_engine2/sdk/EE2Runjob.py

lib/execution_engine2/db/MongoUtil.py
MrCreosote: Hum, would the cancel job endpoint throw an error in that case?

So if I start a job normally and then cancel it, it doesn't affect fair share?

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -57,6 +58,39 @@ def test_init_ok(self):
...
+        job_ids = [job_id1, job_id2]
+        scheduler_ids = [job.scheduler_id, job2.scheduler_id]
+        jobs_to_update = list(map(JobIdPair, job_ids, scheduler_ids))
+        self.getMongoUtil().update_jobs_to_queued(jobs_to_update)
+        job.reload()
+        job2.reload()
+        assert all(j.status == Status.queued.value for j in [job, job2])
+
+    def test_update_jobs_failure(self):
+        job = get_example_job(status=Status.created.value)
```
MrCreosote, line 75: In looking at this test again I think it should be structured a little differently, something like this:

```
for state in [<all states except created>]:
    jobid1 = create_job_with_state(created)
    jobid2 = create_job_with_state(state)
    assertRaises(mongoUtil.update(jobid1, jobid2), Exception("error text"))
```

E.g. individually check that the code will barf on all states except created.

MrCreosote (changes requested):

lib/execution_engine2/db/MongoUtil.py
bio-boris: No, it doesn't throw an error, it just lets the job run and get cancelled when it runs.

Both affect fair share, but the impact is different on having 1 job vs 100 jobs vs 10000 jobs. How much it really matters I am not sure, as in theory they shouldn't run long, but they may end up reserving many cores which could boost the score.  Its possible it doesn't matter that much, but technically the fair share is affected differently.

I see some details at https://htcondor.readthedocs.io/en/latest/admin-manual/user-priorities-negotiation.html?highlight=Real%20User%20Priority%20(RUP)

MrCreosote: So just to be clear, this line won't throw an error if `job.scheduler_id` is `None`?
https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/sdk/EE2Status.py#L126

MrCreosote: Regarding fair share: if starting jobs and then cancelling them - either directly via the Condor API or indirectly via the job polling for the canceled state - affect the fair share state, then for the quick cancel case we're talking about a job that will run an extra X seconds, where X is the poll interval. Is that really enough of an issue to worry about here? Or am I missing something?

bio-boris: Right, it won't throw an error.
```
>>> class Job:
...     apple = None
...
>>> a = Job()
>>> a.apple
>>> print(a.apple)
None
>>> print(f"{a.apple}.0")
None.0
```
https://github.com/kbase/execution_engine2/blob/3150622a20945cf7a28a77f8356098013a340445/lib/execution_engine2/utils/Condor.py#L280

Doesn't allow for exceptions to bubble up


I'm thinking for small amounts of jobs it shouldn't, but for large amounts, it might. I'm not sure how many seconds it takes for starting up a job, but if its like 15-30 seconds before the jobRunner files are transferred to the node, resources are calculated for HTCondor,  it adds up, especially if you have 100 jobs running for 15-30 seconds taking up an entire large memory node.

test/tests_for_db/ee2_MongoUtil_test.py
bio-boris: Alright, updated the test

bio-boris: Actually consolidated the success and failure into one test

lib/execution_engine2/db/MongoUtil.py
bio-boris: But considering that sometimes a single job can run for days and then just die, maybe that's not such a big deal.

test/tests_for_db/ee2_MongoUtil_test.py
MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
MrCreosote: This conversation is getting long but it's important... the level of complexity in starting a job is going to get quite a bit higher with threading.

Ok, so we still have to figure out what to do when a job is canceled before it's updated in mongo (or if the job is in some other terminal state). 

For non-cancel (error, complete, terminated) terminal states, in my opinion it shouldn't throw an error, and should just leave the jobs as is.

For the canceled jobs, I can think of two choices currently:
1) Leave the jobs alone and let the job poll cancellation method do its thing. This might cost the user a few seconds or minutes of fair share time per job.
2) Maybe do a two stage update:
- update the scheduler ID for all the jobs to make sure it's recorded in the DB regardless of the state.
  - this might be worth doing regardless.
- update jobs that are in the created state to updated.
  - if `n_updated > len(jobs)` then pull the jobs, and return the jobs that aren't in the updated or created state.
  - EE2RunJob calls `cancel_job()` on jobs that are in the canceled state.

Any other ideas?

bio-boris: Other idea is that since we know the scheduler ids, we could catch the exception and then cancel the jobs in htcondor with the api.

MrCreosote: Saving the scheduler ID also aids with forensics in the future

MrCreosote: >   catch the exception and then cancel the jobs

You mean cancel all the jobs? If so that seems odd to me - if I cancel a single job quickly all my jobs get cancelled, but if I wait a little bit then just the one job gets canceled?

bio-boris: Yes cancel all the jobs in HTCondor. Why does that seem odd, you then don't have to worry about having the job start up before it realizes it needs to cancel itself.

MrCreosote: It seems odd because I told the UI to cancel one of my ten jobs, but they all wind up getting canceled, assuming I'm understanding you.

bio-boris: Oooh I see.. I didn't think about being able to cancel one of the jobs, I was thinking of cancelling all of the jobs at once. That would be odd..

In that case, then maybe the way to proceed would be to see if the operation to update all jobs didn't equal the number of jobs, and then go through and do the 2nd update
Update * where job id in [ list of ids we just had] and job.scheduler_id is blank, update the scheduler_id (and set to cancel state just to be sure?) and then call HTCondor.cancel on the scheduler_id....

Actually I just realized you can call an HTCondor.cancel using the ee2 id, without knowing the scheduler_id
condor_rm -constraint 'JobBatchName=="ee2_id"'

MrCreosote: Ok, so maybe update cancel_job to use the ee2 id instead of the scheduler ID, and then this is moot?

MrCreosote: Personally I'd still always add the scheduler ID to the mongo record as well because why not

bio-boris: Changed it around, removed the exception. 
I was thinking we could also just mark the record as cancelled, but throw an exception telling the user to wait a few seconds and try to cancel again ?

add
      ```  if job.scheduler_id is None:
            raise CondorJobNotFoundException(f"Couldn't cancel {job_id}. Please try again in a few seconds")
```
to the end of cancel_job

Comment (MrCreosote) [reply lib/execution_engine2/db/MongoUtil.py]: I don't think you pushed the changes?

> I was thinking we could also just mark the record as cancelled, but throw an exception telling the user to wait a few seconds and try to cancel again ?

But then if the front end refreshes the job will show as cancelled, so why would use user try again (even if the front end lets them, which I assume it won't)

Comment (bio-boris) [reply lib/execution_engine2/db/MongoUtil.py]: Ok pushed. 

I believe in the frontend for batch upload, it doesn't show as cancelled, and it handles cancel_job failures by telling you that the cancellation failed.
Now that I think about it, not actually sure what happens to the cancel job button or the entire cell for single jobs when the cancel endpoint raises an exception.

Comment (MrCreosote) [reply lib/execution_engine2/db/MongoUtil.py]: > I believe in the frontend for batch upload, it doesn't show as cancelled,

I don't understand this. If it pulls the job from the backend and the job is in the cancelled state, the front end will ignore that?

Comment (bio-boris) [reply lib/execution_engine2/db/MongoUtil.py]: That would depends on whether or not the next cell state transition depends on what happens after the cancel job is called. If that state transition depends on the successful cancel job api call, it wouldn't transition.

Comment (bio-boris) [reply lib/execution_engine2/db/MongoUtil.py]: Something is in the works right now https://github.com/kbase/narrative/pull/2384

Comment (bio-boris) [reply lib/execution_engine2/db/MongoUtil.py]: I'm talking about the response to the call though, not about the polls. The typical place for the user to cancel a job is in the cell or the job browser, so that's where they want to know about the cancellation failure.

Comment (bio-boris) [reply lib/execution_engine2/db/MongoUtil.py]: We could just add another lookup to see if any of these jobs are now in a terminated state and cancel them if required.

Comment (MrCreosote) [reply lib/execution_engine2/db/MongoUtil.py]: I'm not sure we're understanding each other. If I understand you correctly, what'll happen on a fast cancel is:

* The user starts a batch job and cancels at least one of the jobs quickly enough that condor submission hasn't occurred (which could take seconds to minutes depending on the number of jobs)
* The job is marked as canceled, but not actually canceled
* The user gets a message saying to try and cancel again in a few minutes
* But if the user, say, walks away for a bit and then loads to the job broswer to cancel the job, it'll show as already canceled when it actually isn't.

That seems incorrect to me. Please correct me if I'm wrong about what happens

Comment (bio-boris) [reply lib/execution_engine2/db/MongoUtil.py]: Yes that is mostly correct. The job is marked as cancelled, but that is a "partial" cancel and that prevents the jobRunner from starting and ends the job if it's running once the cancel checker runs. If condor is unavailable, not responding there's not much we can do right now about stopping the job in condor once it comes back without a other reaper or Kafka. The jobRunner would have  to query the job state and make a decision right now.  If condor is available, then we could have this thread check for cancellation status right before it ends, and issue out cancel job commands once more.

Comment (MrCreosote) [reply lib/execution_engine2/db/MongoUtil.py]: Offline we decided that if a user hits cancel before the jobs are submitted to condor:

* the job will be marked as canceled
* the user won't get any abnormal error message, to them it'll look like the job is canceled
* at the end of the job submission thread, the thread will re-pull the jobs from mongo and send cancel messages to Condor for any cancelled jobs, ignoring any errors that may occur due to the job already being cancelled

Worst case this means that the job runs for, at most, the length of the JobRunner cancellation poll interval before stopping.

@bio-boris if this isn't your understanding let's talk again, but I think this is what we decided.

**Comment by bio-boris** [comment]
Working on tests and coverage

File: test/utils_shared/producer.py [hunk 20]
```diff
@@ -22,11 +22,8 @@ def send_kafka_message(self, message, topic=DEFAULT_TOPIC):
             producer = Producer({"bootstrap.servers": self.server_address})
             producer.produce(topic, str(message), callback=_delivery_report)
             producer.poll(2)
-            logging.info(
-                f"Successfully sent message to kafka at topic={topic} message={json.dumps(message)} server_address={self.server_address}"
-            )
         except Exception as e:
-            logging.info(
+            logging.error(
```
bio-boris, line 26: Removed extra info from logs

test/tests_for_sdkmr/ee2_kafka_test.py
```diff
@@ -29,7 +29,7 @@ def setUpClass(cls):
 
     def test_status_change(self):
 
-        with self.assertRaisesRegexp(
+        with self.assertRaisesRegex(
```
bio-boris, line 32: fixing warning

test/tests_for_sdkmr/ee2_SDKMethodRunner_test_EE2Runjob_test.py
```diff
@@ -334,7 +334,7 @@ def test_retry_job_multiple(self, rq_mock, condor_mock):
             "'123' is not a valid ObjectId, it must be a 12-byte input or a 24-character "
             "hex string"
         )
-        with self.assertRaisesRegexp(RetryFailureException, errmsg):
+        with self.assertRaisesRegex(RetryFailureException, errmsg):
```
bio-boris, line 337: fixing warning

test/tests_for_sdkmr/ee2_SDKMethodRunner_test.py
```diff
@@ -30,10 +30,10 @@
     JobRequirementsResolver,
     RequirementsType,
 )
-from lib.execution_engine2.db.models.models import Job, Status, TerminatedCode
-from lib.execution_engine2.exceptions import InvalidStatusTransitionException
-from lib.execution_engine2.sdk.SDKMethodRunner import SDKMethodRunner
-from lib.execution_engine2.utils.CondorTuples import SubmissionInfo
+from execution_engine2.db.models.models import Job, Status, TerminatedCode
```
bio-boris, line 33: fixing imports

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +253,135 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+                    message=KafkaQueueChange(
+                        job_id=job_id,
+                        new_status=Status.queued.value,
+                        previous_status=Status.created.value,  # TODO maybe change this to allow for estimating jobs
+                        scheduler_id=job.scheduler_id,
+                    )
+                )
+            elif job.status == Status.terminated.value:
+                # Remove from the queue, now that the scheduler_id is available
+                self._safe_cancel(job_id, TerminatedCode.terminated_by_user)
```
MrCreosote, line 330: What if the job was terminated for some other reason?

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -3,18 +3,20 @@
 import time
 import traceback
 from contextlib import contextmanager
-from typing import Dict
+from datetime import datetime
+from typing import Dict, List, Tuple
```
MrCreosote, line 7: Tuple unused

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -2,12 +2,13 @@
...
 import unittest
-from configparser import ConfigParser
 
 from bson.objectid import ObjectId
 
 from execution_engine2.db.MongoUtil import MongoUtil
-from execution_engine2.db.models.models import Job, JobLog
+from execution_engine2.db.models.models import Job, JobLog, Status
+from execution_engine2.sdk.EE2Runjob import JobIdPair
+from execution_engine2.exceptions import InvalidStatusTransitionException
```
MrCreosote, line 11: unused

MrCreosote (changes requested):
code seems fine but the tests aren't passing, so I assume there's more work there to do and haven't looked at them yet

lib/execution_engine2/db/MongoUtil.py
MrCreosote: :+1:

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Nothing actually happens since
```
    def cancel_job(self, job_id=None, terminated_code=None):
            j = self.get_job(job_id)
            if self.check_if_already_finished(j.status):
                return False
```

lib/execution_engine2/db/MongoUtil.py
bio-boris: fixed

test/tests_for_db/ee2_MongoUtil_test.py
bio-boris: fixed

bio-boris:
Tests are done, not sure why that one failed all of a sudden on github actions. It's been failing locally on my machine for months but not on github actions. It was a bad test because the underlying functionality is actually wrong and the test is flaky depending on where you run it

MrCreosote:
Interesting, the tests passed consistently on my machine when I was doing ee2 work.

lib/execution_engine2/db/MongoUtil.py
MrCreosote: :+1:

test/tests_for_db/ee2_MongoUtil_test.py
MrCreosote: :+1:

MrCreosote:
Note to self: New behaviors are
1) Batch jobs are saved via SDKMR.save_jobs vs save_job (calling MongoUtil.insert_jobs)
2) Jobs are set to updated by MongoUtil.update_jobs_to_queued
3) Any jobs canceled prior to the update in 2) have SDKMR.cancel_job called on them
4) Parent job is updated with child jobs via SDKMR.add_child_jobs

MrCreosote:
Further note to self: there are API to DB integration tests that exercise `run_batch` which pass without changes, which is reassuring

MrCreosote (changes requested):
Looks good, just a few minor changes and some unit tests that I think should be pretty easy

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +267,68 @@ def check_if_already_finished(job_status):
...
+                )
+
+            bulk_update_scheduler_jobs.append(
+                UpdateOne(
+                    {
+                        "_id": ObjectId(job_id_pair.job_id),
+                    },
+                    {
+                        "$set": {
+                            "queued": queue_time_now,
```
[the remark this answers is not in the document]
bio-boris: Right, but `test_run_job_batch` calls `_get_mongo_job` and `assert_close_to_now(job.pop("updated"))` but I can add something similar in here. Done

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -420,6 +486,18 @@ def update_job_status(self, job_id, status, msg=None, error_message=None):
     def mongo_engine_connection(self):
         yield self.me_connection
 
+    def insert_jobs(self, jobs_to_insert: List[Job]) -> List[ObjectId]:
+        """
```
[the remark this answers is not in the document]
bio-boris: Ok, wrote a basic test

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -263,6 +267,68 @@ def check_if_already_finished(job_status):
...
+            bulk_update_scheduler_jobs.append(
+                UpdateOne(
+                    {
+                        "_id": ObjectId(job_id_pair.job_id),
+                    },
+                    {
+                        "$set": {
+                            "queued": queue_time_now,
+                            "scheduler_id": job_id_pair.scheduler_id,
+                            "scheduler_type": scheduler_type,
```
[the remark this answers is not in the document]
bio-boris: ok added some

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -56,8 +62,8 @@
 _USER = "someuser"
 _TOKEN = "tokentokentoken"
 _OTHER_USER = "some_sucker"
-_CREATED_STATE = "created"
-_QUEUED_STATE = "queued"
+_CREATED_STATE = Status.created.value
+_QUEUED_STATE = Status.queued.value
```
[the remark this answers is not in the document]
bio-boris: I thought it was weird to think of checking these strings directly. In that case, doesn't that mean we should also be checking to make sure  that the strings we are hardcoding in here  pass the validity checking functions in models.py like `valid_status` `valid_authstrat` `valid_termination_code` and `valid_errorcode`  ?

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -184,7 +189,9 @@ def _set_up_common_return_values(mocks):
     mocks[MongoUtil].get_job.return_value = retjob
 
 
-def _check_common_mock_calls(mocks, reqs, wsid, app=_APP, parent_job_id=None):
+def _check_common_mock_calls(
+    mocks, reqs, wsid, app=_APP, parent_job_id=None, batch=False
```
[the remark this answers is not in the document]
bio-boris: removed

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -763,7 +769,7 @@ def _run_and_run_batch_fail(
     _run_batch_fail(rj, [params], {}, as_admin, expected)
 
 
-def _set_up_common_return_values_batch(mocks):
+def _set_up_common_return_values_batch(mocks, returned_job_state=None):
```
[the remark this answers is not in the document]
bio-boris: Good idea

test/tests_for_sdkmr/EE2Runjob_test.py
```diff
@@ -882,38 +912,51 @@ def _check_common_mock_calls_batch(mocks, reqs1, reqs2, parent_wsid):
...
-                )
-            ),
-        ]
-    )
+    child_job_pairs = [
+        JobIdPair(_JOB_ID_1, _CLUSTER_1),
+        JobIdPair(_JOB_ID_2, _CLUSTER_2),
+    ]
+    mocks[MongoUtil].update_jobs_to_queued.assert_has_calls([call(child_job_pairs)])
+
```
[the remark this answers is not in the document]
bio-boris: added

lib/execution_engine2/sdk/SDKMethodRunner.py
```diff
@@ -251,13 +252,28 @@ def check_as_concierge(self):
     # at this point since MongoEngine creates a global connection to MongoDB
     # and makes it available to all the model objects.
 
+    def save_jobs(self, jobs: List[Job]) -> List[str]:
+        """
```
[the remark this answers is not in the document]
bio-boris: added some

lib/execution_engine2/sdk/SDKMethodRunner.py
```diff
@@ -251,13 +252,28 @@ def check_as_concierge(self):
...
+        return [str(job_id) for job_id in job_ids]
+
     def save_job(self, job: Job) -> str:
         """
         Save a job record to the Mongo database and return the job's ID as a string.
         """
         job.save()
         return str(job.id)
 
+    def add_child_jobs(self, batch_job: Job, child_jobs: List[str]):
```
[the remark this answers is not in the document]
bio-boris: added

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -258,14 +253,137 @@ def _prepare_to_run(self, params, concierge_params=None) -> JobSubmissionParamet
...
+                        new_status=Status.queued.value,
+                        previous_status=Status.created.value,  # TODO maybe change this to allow for estimating jobs
+                        scheduler_id=job.scheduler_id,
+                    )
+                )
+            elif job.status == Status.terminated.value:
+                # Remove from the queue, now that the scheduler_id is available
+                # The job record doesn't actually get updated in the db a 2nd time, and this TerminatedCode is only
+                # used by the initial transition to Terminated
+                self._safe_cancel(job_id, TerminatedCode.terminated_by_user)
```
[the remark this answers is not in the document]
bio-boris: Are you saying its too hard to do it, or that we need to do it?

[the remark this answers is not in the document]
MrCreosote: I'm wondering if you can think of a reasonable way to exercise this in an integration test (I can't).

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
MrCreosote: We do that in the `api_to_db_test`, right?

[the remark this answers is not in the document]
MrCreosote: this is a unit test just for EE2RunJob, so I don't think so. It really shouldn't care about how other modules handle the strings - just that the strings are part of the expected output from the method calls

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
MrCreosote: Yeah, I'm always going to ask for unit tests when it's reasonably possible to do since they're usually a lot easier to understand and debug than integration tests.

Anyway :+1:

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
MrCreosote: :+1:

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
MrCreosote: :+1:

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
MrCreosote: :+1:

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
MrCreosote: It tests that the method returns job IDs, but not that the jobs are actually correctly inserted to the DB. If this were me writing the tests I'd retrieve the jobs either via the MongoUtil API or directly from mongo and ensure the data was what I expected

lib/execution_engine2/sdk/SDKMethodRunner.py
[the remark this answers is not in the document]
MrCreosote: This test doesn't seem to actually test anything? https://github.com/kbase/execution_engine2/pull/406/files#diff-d7d713ca6587e357238406d58fce2bf6e82ddb5339eb51dfafeaf6699103e3bfR177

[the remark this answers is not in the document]
MrCreosote: The test that does run looks good though

lib/execution_engine2/sdk/SDKMethodRunner.py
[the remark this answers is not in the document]
MrCreosote: Test looks good other than it doesn't check the return value

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
bio-boris: Mock out `self.sdkmr.get_mongo_util().get_jobs(job_ids)` to return jobs that have a terminated code?

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
bio-boris: The source of truth for these values should be from the Status enum, so not using it seems very wrong to me. But ok, will change it back.

lib/execution_engine2/sdk/SDKMethodRunner.py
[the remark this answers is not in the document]
bio-boris: Added

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
bio-boris: Alright, added that, but not quite sure if that's what you had in mind

lib/execution_engine2/sdk/SDKMethodRunner.py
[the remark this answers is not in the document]
bio-boris: Oops deleted that one

lib/execution_engine2/sdk/EE2Runjob.py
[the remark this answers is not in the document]
MrCreosote: yeah, it's not really an integration test in that case. Ok, forget it, I honestly thought it was probably going to be too hard

lib/execution_engine2/sdk/SDKMethodRunner.py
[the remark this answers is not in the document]
MrCreosote: :+1:

lib/execution_engine2/sdk/SDKMethodRunner.py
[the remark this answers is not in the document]
MrCreosote: :+1:

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
MrCreosote: it is, except the problem is that equality on Job objects only tests the `_id`, not any other fields. So the insert method could completely trash the incoming Job object and save it and the test will still pass. I'd update to use the `assert_job_equal` method from `EE2RunJob_test.py` (maybe that should be moved to a utility class at some point)

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
MrCreosote: > following this logic we can make these strings to be "cat" or "dog" within the test if we'd like

If we did that the tests would fail right? The states are set in `_init_job_rec` and `update_job_to_queued`. When the `save_job.assert_called_once_with` or the equivalent for update is called in the tests, those would fail if the state strings are wrong.

That's the point of not using constructs from the code under test in the test - if those constructs change then the tests fail, which is what we want. 

So the source of truth is still the `Status` values and these tests ensure the the source of truth is not changed without being a deliberate action on the part of the coder.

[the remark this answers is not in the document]
MrCreosote: anyway, fixed so :+1:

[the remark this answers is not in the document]
bio-boris: So why would we use these tests to ensure that the source of truth is not changed, rather than a test dedicated to checking the source of truth?

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
bio-boris: Alright well how about using
```
        for i, retrieved_job in enumerate(retrieved_jobs):
            assert jobs_to_insert[i].to_json() == retrieved_job.to_json()
            assert jobs_to_insert[i].to_mongo() == retrieved_job.to_mongo()
            assert jobs_to_insert[i].to_dbref() == retrieved_job.to_dbref()
```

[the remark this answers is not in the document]
MrCreosote: Seems reasonable

test/tests_for_sdkmr/EE2Runjob_test.py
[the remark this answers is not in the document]
MrCreosote: You could certainly have both. These tests ensure that the source of truth is used correctly in EE2RunJob though, which a test that only checks the contents of the source of truth wouldn't do.

MrCreosote (approved):
LGTM

lib/execution_engine2/db/MongoUtil.py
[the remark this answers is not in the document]
bio-boris: Bah, realized that the modify command isn't updated the "updating" timestamp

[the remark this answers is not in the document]
bio-boris: Added this test in a new PR #416

PR #414, opened 2021-08-06
DATAUP-555 Return mappings

bio-boris:
# Description of PR purpose/changes

* Please include a summary of the change and which issue is fixed. 
* Please also include relevant motivation and context.
* List any dependencies that are required for this change.

# Jira Ticket / Github Issue #
- [ ] Added the Jira Ticket to the title of the PR e.g. (DATAUP-69 Adds a PR template)

# Testing Instructions
* Details for how to test the PR: 
- [ ] Tests pass in Github Actions and locally 
- [ ] Changes available by spinning up a local test suite and doing X

# Dev Checklist:

- [ ] My code follows the guidelines at https://sites.google.com/truss.works/kbasetruss/data-upload-project/development
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have run Black and Flake8 on changed Python Code manually or with git precommit (and the Github Actions build passes)

# Updating Version and Release Notes (if applicable)

- [ ] [Version has been bumped](https://semver.org/) for each release
- [ ] [Release notes](/RELEASE_NOTES.md) have been updated for each release (and during the merge of feature branches)

PR #415, opened 2021-08-06
DATAUP-555 Dedupe and fail

bio-boris:

* Allow job retry requests to error out and complete in one api call

# Testing Instructions

# Dev Checklist:


ialarmedalien (approved):
this is perfect!

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -647,44 +647,38 @@ def retry_multiple(
...
-                jobs.append(job)
-                batch_jobs.append(batch_job)
             except Exception as e:
-                raise RetryFailureException(e)
-
-        # Submit all of the collected jobs
-        for i, job_id in enumerate(job_ids):
+                # Collect the presubmit error and don't submit the job
+                retried_jobs.append({"job_id": job_id, "error": f"{e}"})
+                continue
```
MrCreosote, line 668: Would it make sense to log the stack trace here? We're completely losing the trace IIUC

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -647,44 +647,38 @@ def retry_multiple(
...
+            # Presubmit worked, write to the db and submit
```
MrCreosote, line 669: the way this works, IIUC, is that you can have both errors and successful retries in the same response. I understood from the Slack thread that the desired behavior was that there should only be errors or successful retries in the response, not both.

MrCreosote (changes requested):

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: Depends if we can easily fix the error without the stack trace. If we can, we don't need it, if we need the stack trace, then we should include it. Do you need the stacktrace @ialarmedalien ?

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: This is the desired behavior now. We closed https://github.com/kbase/execution_engine2/pull/414 because that was the behavior we didn't want anymore.

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: But the except catches all exceptions. How can you know if we can fix the error for any possible error that can occur?

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: That doesn't appear to be the case based on my reading of the thread: https://kbase.slack.com/archives/C01T9GG43C4/p1628286203069900?thread_ts=1628274397.052200&cid=C01T9GG43C4

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: You don't, but if we can fix it, then we can, and we cannot, then we cannot. (We don't even document the types of possible exceptions that can be returned by the endpoint, and if we did, we would have to parse the string for errors?   )

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: That seems to be the desired behavior to me, and the PR is already approved with this behavior too, so that seems to be the case for @ialarmedalien too

MrCreosote: The approval is prior to the slack comment I reference above

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: > You don't, but if we can fix it, then we can, and we cannot, then we cannot. 

how can you know if you can fix it or not without the stack trace? The stack trace may give you the information you need to diagnose or fix the error, especially for something like a `KeyError`.

> We don't even document the types of possible exceptions that can be returned by the endpoint

I'm confused here - what does this have to do with logging the exception?

bio-boris: >how can you know if you can fix it or not without the stack trace? The stack trace may give you the information you need to diagnose or fix the error, especially for something like a KeyError.
Right, it would be hard for something like a KeyError

However, If the error message for the thrown exception is descriptive enough that you don't need a traceback, then you can fix it without the traceback. If it seems that most errors won't be descriptive enough then we may need a traceback.  There's plenty of APIs that return an error without a traceback. But maybe since our jsonRPC endpoints don't, we need the traceback for most of the errors?

>I'm confused here - what does this have to do with logging the exception?

It doesn't have to do with logging the exception, it has to do with fixing the exception as per your comment
> How can you know if we can fix the error for any possible error that can occur?

You can't know if you can fix the error or not. If its a system error, you can't fix it, and if its a user error, you might be able to fix it.

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: @ialarmedalien Can you comment?

lib/execution_engine2/sdk/EE2Runjob.py
MrCreosote: > There's plenty of APIs that return an error without a traceback.

Hold on - are you talking about adding the traceback to the API return? I'm talking about logging it, just
```
self.logger.error(e)
```

bio-boris: Yea, for some reason I thought by log you meant include in the API return, which makes sense to me if our errors messages are so bad that we don't know what the error is without the traceback.

MrCreosote: > Yea, for some reason I thought by log you meant include in the API return

Nopety nope nope

> our errors messages are so bad that we don't know what the error is without the traceback.

It's not even so much that, it's that when something unexpected happens or we have a bug (like a KeyError) then we're not writing the error messages, and so we're at the mercy of the authors of the code where it originated.

Also having line numbers both for our code and 3rd party code makes it much faster to fix bugs.

bio-boris: I'd rather not pollute the logs with logging the (first exception) presubmit error, but writing to the db and submitting probably might be good candidates for logging (second exception).  Do you have strong feelings about logging both of them?

bio-boris: Oh I see. Maybe we should probably have a general exception handler code that logs certain errors to the logs, and doesn't log others? And we should log the errors right when they occur, and not wait till they bubble up to log them?

MrCreosote: > I'd rather not pollute the logs with logging the (first exception) presubmit error, but writing to the db and submitting probably might be good candidates for logging (second exception). Do you have strong feelings about logging both of them?

The issue I have with not logging the 1st error is that if there's a bug in that code then it's going to make debugging stuff much harder. If the try / catch only caught errors where we know we don't need the exception that'd be fine, but right now it catches everything. Once potential option is making a custom exception subclass for the types of errors you care about and just catch that.

MrCreosote: > Oh I see. Maybe we should probably have a general exception handler logger code that logs certain errors to the logs, and doesn't log others? And we should log the errors right when they occur, and not wait till they bubble up to log them?

That all sounds fine to me. As long as we're not losing stack traces we need to debug issues in the field I'm happy.

MrCreosote: For now we could just log everything and add the improved code that's smarter about what gets logged later

bio-boris: Hmm. Does that mean we would have to change every ValueError and bare exception we raise to be a subclass of ExecutionEngineException , and check for ExecutionEngineExceptions vs regular exceptions?

bio-boris: I've been removing logging code, because I got tired of looking through logs that were useless so I'm hesitant to add more logging code but if you think thats what we need we can add in logger.error or logger.debug statements.

MrCreosote: > Does that mean we would have to change every ValueError ...

yep, just one suggestion, there may be other ways to do this

MrCreosote: > I've been removing logging code, because I got tired of looking through logs that were useless so I'm hesitant to add more logging code but if you think thats what we need we can add in logger.error or logger.debug statements.

yeah, I hear you there, but when ee2 starts throwing weird unexpected exceptions we'll start cursing if there's no way to get the exceptions.

lib/execution_engine2/sdk/EE2Runjob.py
```diff
@@ -762,44 +762,38 @@ def retry_multiple(
...
-            if job_ids.count(job_id) > 1:
-                offending_ids[job_id] += 1
-
-        if offending_ids.keys():
+        offending_ids = [item for item, count in Counter(job_ids).items() if count > 1]
+        if offending_ids:
             raise ValueError(
                 f"Retry of the same id in the same request is not supported."
-                f" Offending ids:{list(offending_ids.keys())} "
+                f" Offending ids:{offending_ids} "
```
ialarmedalien, line 769: put in a space before `{offending_ids}`

lib/execution_engine2/sdk/EE2Runjob.py
ialarmedalien: This is the desired behaviour: we collect results for every item in the input set, so we could have all successful retries, all errors, or a mixture of both. The important thing is that there is a result for every input item.

MrCreosote: Ok, it seems that I misinterpreted what AJ said. I still think that implementing partial writes is a mistake and makes it more difficult for the caller, but if both you and AJ want partial writes I'll acquiesce

lib/execution_engine2/sdk/EE2Runjob.py
bio-boris: added

PR #416, opened 2021-08-06
Fix modify and save test

bio-boris:


MrCreosote:
Did I miss something in my review or is this a pre-existing issue?

bio-boris:
This is a pre-existing issue that I noticed during https://github.com/kbase/execution_engine2/pull/406#discussion_r684508289

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -58,17 +58,47 @@ def test_init_ok(self):
...
+            retrieved_job_json = json.loads(retrieved_job.to_json())
+            retrieved_job_mongo = retrieved_job.to_mongo()
+
+            # Cannot compare `id` or `updated` for a job that wasn't yet saved
+            if "_id" not in original_job_json:
+                del retrieved_job_json["_id"]
+                del retrieved_job_mongo["_id"]
+            if "updated" not in original_job_json:
+                del retrieved_job_json["updated"]
+                del retrieved_job_mongo["updated"]
```
MrCreosote, line 98: Aren't these two `if`s going to be always true or always false?

bio-boris: Fixed

lib/execution_engine2/db/models/models.py
```diff
@@ -339,9 +339,13 @@ class Job(Document):
     meta = {"collection": "ee2_jobs"}
 
     def save(self, *args, **kwargs):
-        self.updated = time.time()
+        self.updated = datetime.utcnow().timestamp()
         return super(Job, self).save(*args, **kwargs)
 
+    def modify(self, query=None, **update):
+        update["updated"] = datetime.utcnow().timestamp()
+        return super(Job, self).modify(query=None, **update)
```
MrCreosote, line 347: Shouldn't this just pass the query as is, not pass `None`?

test/tests_for_db/ee2_MongoUtil_test.py
```diff
@@ -58,17 +58,55 @@ def test_init_ok(self):
         mongo_util = self.getMongoUtil()
         self.assertTrue(set(class_attri) <= set(mongo_util.__dict__.keys()))
 
+    def test_save_and_modify_updates_timestamp(self):
+        job = get_example_job(status=Status.created.value)
+        updated_ts = job.updated
+        job.save()
```
MrCreosote, line 64: Its seems a bit odd that saving a job updates the timestamp even if there are no changes, is that what you want?

lib/execution_engine2/db/models/models.py
MrCreosote: Assuming I understand this I guess no code uses the query field otherwise tests should've broken

MrCreosote (commented):
Couple of questions

lib/execution_engine2/db/models/models.py
bio-boris: Yea it should pass the query good call

test/tests_for_db/ee2_MongoUtil_test.py
bio-boris: Well, updated tracks the last time the job was saved or modified, so seems ok to me. What do you suggest?

lib/execution_engine2/db/models/models.py
MrCreosote: how did the tests pass with this bug?

test/tests_for_db/ee2_MongoUtil_test.py
MrCreosote: I mean updated to me means that the job was updated, e.g. changed in some way. It doesn't make sense to me for the updated timestamp to change if the job hasn't changed

bio-boris: Alright, so if we make sure to only save jobs with changes, then we should be alright?

bio-boris: I'm not sure if its possible to know if a record has a different value or not when saving unless we reload the record or do two saves?

MrCreosote: > Alright, so if we make sure to only save jobs with changes, then we should be alright?

I think so, but we should document the behavior clearly

MrCreosote: > I'm not sure if its possible to know if a record has a different value or not when saving unless we reload the record or do two saves?

You can do a query on the entire record, but that's probably overkill. It'd be nice if Mongo had a feature where a field would only be updated if at least one of a set of provided fields changed

PR #422, opened 2021-09-22
DATAUP-262 Create ADR For Split/Aggregate Jobs

bio-boris:
* ADR

### DIFF
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -22,7 +22,7 @@ In order to do this, some apps use a mechanism called KBParallel. The apps are l
 
 * Current UI is not adequate: Users shouldn’t have to code in order to run batch analysis. Also it’s difficult to do so, even for those familiar with KBase code (have to find object names)
 * Dependency on [KBParallel](https://github.com/kbaseapps/KBParallel): any changes to KBParallel could affect KB Batch and subsequently all other apps. 
-* Queue deadlocking: users have a max of 10 slots in the queue, with the current implementation one is taken up just to manage the jobs. Could lead to deadlock scenarios
+* Queue deadlocking: users have a max of 10 slots in the queue, with the current implementation one management job is created to manage the jobs that it submits. This could lead to deadlock scenarios, as there can be 10 management jobs waiting to submit computation jobs, but they cannot, as there all slots are being used up.
 * KBP can spawn other KBP jobs. Batch jobs can spawn other batch jobs. 
 * Missing the ability to be able to run, manage (cancel) and track jobs and their subjobs along with the ability to specify resources differently between the main and sub jobs
 * No good way to test and hard to benchmark or measure performance 
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -125,11 +125,20 @@ Still to be determined (not in scope of this ADR):
 * `+` Simplest solution, quickest turnaround, fixes deadlock issue
 * `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs 
 
-### Increase number of slots or Seperate Queue for kbparallels apps without 10 job limit
-* `+` Simple solutions, quick turnarounds, fixes deadlock issue
+### Increase number of slots per user > 10
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue for small numbers of jobs.
+* `-` Doesn't fix deadlock issue as the user can still submit more KBP jobs
 * `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
 * `-` A small amount of users can take over the entire system
 * `-` The calculations done by the apps will interfere with other apps and cause crashes/failures
+
+###  Seperate Queue for kbparallels apps that may or may not have its own limit to running jobs.
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue 
+* `+` Requires minimum changes to ee2 and condor if condor supports this feature
+* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
+* `-` A small amount of users can take over the entire system unless  the new queue has its own limit to running jobs then it prevents users from taking over.
+* `-` The calculations done by the apps will interfere with other apps and cause crashes/failures
+
 
 ### Modify KBP to do only local submission, Move the job to a machine with larger resources
 * `+` Simple solutions, quick turnarounds, fixes deadlock issue, fixes UI issues
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -130,15 +130,19 @@ Still to be determined (not in scope of this ADR):
 * `-` Doesn't fix deadlock issue as the user can still submit more KBP jobs
 * `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
 * `-` A small amount of users can take over the entire system by being able to submit more than 10 jobs
-* `-` > 10 nodes will be taken up by jobs that do little computation as each job gets its own node
+* `-` > 10 nodes will continue be taken up by jobs that do little computation as each job gets its own node
 
-###  Seperate Queue for kbparallels apps that allows multiple instances of apps that use kbparallels running on the same machine
-* `+` Simple solutions, quick turnarounds, fixes deadlock issue , requires changes to ee2 using an allowlist and condor to make jobs in the new queue a consumable resource.
-* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
-* `-` A small amount of users can take over the new queue unless the new queue has its own limit to running jobs then it prevents users from taking over.
-* `-` The calculations done by the apps will interfere with other apps in the new queue still and cause crashes/failures since they are doing more than managing the jobs right now. This would be solved by not allowing multple instances of apps that use KBP, but then it would waste a lot of resources.
+### LIMIT KBP jobs to a maximum of 5 active jobs per user
+* `+` Simple solution requires ee2 to maintain list of KBP apps, and add a KBP_LIMIT to jobs from this list. [Condor](https://github.com/kbase/condor/pull/26) will need KBP_LIMIT Added
+* `+` List of apps is not frequently updated
+* `-` If a new app uses KBP and their app is not on the list, it won't be limited by the KBP_LIMIT unless the owner lets us know.
+* `-` If an existing app no longer uses KBP, their app is still limited unless the owner lets us know.
+* `-` Nodes will continue be taken up by jobs that do little computation as each job gets its own node.
 
-
+###  LIMIT KBP jobs to a maximum of 5 active jobs per user + Seperate queue for kbparallels apps 
+* `+` Allows us to group up KBP jobs onto fewer machines, instead of giving them their entire node
+* `-` Requires going through each app and understanding the worst case computational needs in order to  set the estimated cpu and memory needs for each app 
+* `-` Apps can interfere with other innocent apps and take them down
 
 ### Modify KBP to do only local submission, Move the job to a machine with larger resources
 * `+` Simple solutions, quick turnarounds, fixes deadlock issue, fixes UI issues
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -117,13 +117,14 @@ Still to be determined (not in scope of this ADR):
 * Writing new ee2 endpoints to entirely handle batch execution and possibly use a DAG
 * Remove kbparallels and change apps to a collection of 2-3 apps that do submit, split and aggregate and an use an ee2 endpoint to create a DAG
 * Different DevOps solutions
+* Rewriting KBP or swapping it out for a lightweight alternative that has a subset of the KBP features
 
 ## Pros and Cons of the Alternatives
 
-### Limit multiple instances of kbparallels by keeping a list of apps that use it
-
-* `+` Simplest solution, quickest turnaround, fixes deadlock issue
-* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs 
+### General Notes
+*  With the current implementation of KBP, Having a separate KBP queue with multiple machines can save a spot from a user's 10 job maximum for running more jobs, but takes up / wastes compute resources (especially if the nodes sit idle). The user still gets  10 jobs, but there are less spots for jobs to run overall in the system if we make another queue, as this requires taking up more compute nodes that are currently dedicated to the NJS queue.
+* Without changing apps that use KBP, running multiple KBP apps on the same machine can interfere with each other and we want to avoid this. 
+* If we scrap KBP in favor of a "lightweight alternative" we can avoid some of the previous issues, if we modify all apps that use KBP to use a lightweight alternative.  A lightweight alternative would have to guarantee that no computation besides job management occured, and then we could have the management jobs sit and wait for other jobs without interfering with other jobs on the system. 
 
 ### Increase number of slots per user > 10
 * `+` Simple solutions, quick turnarounds, fixes deadlock issue for small numbers of jobs.
@@ -131,24 +132,58 @@ Still to be determined (not in scope of this ADR):
 * `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
 * `-` A small amount of users can take over the entire system by being able to submit more than 10 jobs
 * `-` > 10 nodes will continue be taken up by jobs that do little computation as each job gets its own node
+* `-` Capacity is still wasted, as some KBP jobs sit around waiting for other jobs to run
 
-### LIMIT KBP jobs to a maximum of 5 active jobs per user
+### LIMIT KBP jobs to a maximum of N<10 active KBP jobs per user
 * `+` Simple solution requires ee2 to maintain list of KBP apps, and add a KBP_LIMIT to jobs from this list. [Condor](https://github.com/kbase/condor/pull/26) will need KBP_LIMIT Added
 * `+` List of apps is not frequently updated
+* `+` Apps do not need to be modified
 * `-` If a new app uses KBP and their app is not on the list, it won't be limited by the KBP_LIMIT unless the owner lets us know.
 * `-` If an existing app no longer uses KBP, their app is still limited unless the owner lets us know.
 * `-` Nodes will continue be taken up by jobs that do little computation as each job gets its own node.
+* `-` Users may not be able to effectively use up their 10 job spots 
+* `-` Capacity is still wasted, as some KBP jobs sit around waiting for other jobs to run
 
-###  LIMIT KBP jobs to a maximum of 5 active jobs per user + Seperate queue for kbparallels apps 
+###  LIMIT KBP jobs to a maximum of N<10 active jobs per user + Seperate queue for kbparallels apps 
+* `+` Same pros as above
 * `+` Allows us to group up KBP jobs onto fewer machines, instead of giving them their entire node
 * `-` Requires going through each app and understanding the worst case computational needs in order to  set the estimated cpu and memory needs for each app 
 * `-` Apps can interfere with other innocent apps and take them down
 * `-` Creating a new queue requires balancing between how many active KBP nodes there vs how many nodes are available for other NJS jobs.
+* `-` Capacity is still wasted, as some KBP jobs sit around waiting for other jobs to run
+
+
+### Build KBP Lightweight Version + KBP Queue
+
+
+#### Design of new verison
+* All apps must be modified to use the new KBP lightweight version, which will: 
+* Can either modify KBP, or create a new tool/package to use instead of KBP
+
+1) Launch a management job called the *Job Manager* that sits in the KBP Queue, alongside other KBP jobs. 
+2) Launch a new NJS queue job  called the *Setup Job* which will
+ 1) Use the User Parameters and/or
+ 2) Download the Data from the initial parameters  (Optional)
+ 3) The results of information gathered from the initial download and or paramters will be sent to (Job Manager)
+3) The *Job Manager* now has enough parameters to setup *Fan Out* Jobs, and will send out jobs and wait for them (and possibly retry them upon failure)
+4) Fan Out jobs download data and perform calculations, save them back to the system, and return references to the saved objects
+5) The *Job Manager* optionally launches a *Group/Reduce* job based on User Parameters and or the results of *Fan Out* Jobs
+6) The *Group/Reduce* job downloads objects from the system, and creates a set or other grouping, and then saves the set back to the system
+7) The *Job Manager*  sends the results back to a *Report* Job, which downloads the final data and uploads a report based on that data
+8) The *Job Manager* returns the reference to the results of the *Report Job*
+
+Pros/Cons
+* `-` Addresses the deadlocking issue, UI still broken for regular runs and batch runs if we re-use KBP
+* `+` All KBP jobs can run on a small subset of machines
+* `+` Job Manager 
+* `+` Minimal changes to ee2 required
+
  
-### Modify KBP to do only local submission, Move the job to a machine with larger resources
+### Modify Apps to do only local submission by remove KBP, and moving the job 
 * `+` Simple solutions, quick turnarounds, fixes deadlock issue, fixes UI issues
 * `-` We have a limited number of larger resources machines
 * `-` Continued dependency on deprecated KBP tools
+* `-` App runs may take longer since fewer resources may be available to the app run
 
 ### Deprecate kbparallels, and write new ee2 endpoints to entirely handle split and aggregate
 * `+` No longer uses an app, No longer uses a slot in the queue
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -17,7 +17,7 @@ In order to do this, some apps use a mechanism called KBParallel. The apps are l
 * Missing the ability to be able to run, manage and track jobs and their subjobs 
 * No good way to test and hard to benchmark or measure performance 
 * Code is split more than is necessary
-
+* UI doesn't properly display progress of batch jobs
 
 Background:
 
@@ -31,14 +31,13 @@ Pending
 
 ## Alternatives Considered
 
-* Writing a new Batch App
-* Implementing a Dynamic Service
-* Implementing a standalone Narrative Widget
-* Implementing a new (set) of EE2 endpoint(s)
+* Ignore most issues and just make apps that run kbparallels limited to N instances of kbparallels per user to avoid deadlocks
+* Writing new ee2 endpoints to entirely handle batch execution
+* Remove kbparallels and change apps to a collection of 2-3 apps that do submit, split, aggregate and a batch endpoint
 
 ## Decision Outcome
 
-After much [discussion](https://docs.google.com/document/d/1PoiOas-hqgHONNzmCVupjmNCyvUvcned5YfZf-5CxsI/edit#) [and research](https://docs.google.com/spreadsheets/d/1FGecELaEBAWQ7ljXsR29h5RdtmYV_jMyLSiJAWcm4mE/edit#gid=492998772) the decision was to create a new endpoint (or set of endpoints) in EE2 for the purpose of running batch analysis workflows at KBase. 
+
 
 ## Consequences
 
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -16,7 +16,14 @@ Still to be determined (not in scope of this ADR):
 ## Intro
 Sometimes a calculation requires too many resources from one node (walltime, memory, disk), so the calculation gets spread across multiple machines.
 The final step of the app that uses KBParallels is to create a report. This step may use results from all of the computed jobs to create the final report.
-In order to do this, some apps use a mechanism called KBParallel. The apps are listed at [DATAUP-111](https://kbase-jira.atlassian.net/browse/DATAUP-111)
+In order to do this, some apps use a mechanism called KBParallel. The apps listed at [git](https://github.com/search?q=kbparallel&type=code) all use kbparallels. There are 7 known apps 
+* kb_Bowtie2
+* refseq_importer
+* kb_concoct
+* kb_phylogenomics
+* kb_hisat2
+* kb_meta_decoder
+*  kb_Bwa. 
 
 ## The current implementation of Batch Analysis in [kb_BatchApp](https://github.com/kbaseapps/kb_BatchApp) at KBase has the following issues:
 
--- a/docs/adrs/004-SplitAndAggregate.md
+++ b/docs/adrs/004-SplitAndAggregate.md
@@ -4,6 +4,10 @@ Date: 2021-09-22
 
 [Related ADR](https://github.com/kbase/execution_engine2/blob/develop/docs/adrs/rework-batch-analysis-architecture.md)
 
+## Note
+This ADR is more of a place to keep the current discussions we had at https://docs.google.com/document/d/1AWjayMoqCoGkpO9-tjXxEvO40yYnFtcECbdne5vTURo
+Rather than to make a decision. There is still more planning, scoping and testing involved before we can fully design this system.
+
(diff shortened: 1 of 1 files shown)


docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,139 @@
...
+
+Still to be determined (not in scope of this ADR):
+* UI and how it relates to the bulk execution
+* XSV Analysis and how it relates to the bulk execution
+
+
+## Intro
+Sometimes a calculation requires too many resources from one node (walltime, memory, disk), so the calculation gets spread across multiple machines.
+The final step of the app that uses KBParallels is to create a report. This step may use results from all of the computed jobs to create the final report.
+In order to do this, some apps use a mechanism called KBParallel. The apps are listed at [DATAUP-111](https://kbase-jira.atlassian.net/browse/DATAUP-111)
```
MrCreosote, line 19: This list only includes in scope apps for BU, whereas this ADR is really about if we need to support apps outside that list that do job batching

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,139 @@
...
+* Different DevOps solutions
+
+## Pros and Cons of the Alternatives
+
+### Limit multiple instances of kbparallels
+
+* `+` Simplest solution, quickest turnaround, fixes deadlock issue
+* `-` UI still broken for batch analysis
+
+## DEADLOCK: Increase number of slots or Seperate Queue for kbparallels apps without 10 job limit
```
MrCreosote, line 121: I'm starting to think a separate queue with barely any resources for the monitoring job might work. Basically the only responsibility of the jobs in that queue are to start other jobs and monitor them. So the monitor jobs have basically no resources, and all they do is:

1) Kick off a FO job that starts the PIP jobs. If that's not computationally intensive it could be done in the monitor job. The FO job returns the job IDs to the monitor job.
2) When the PIP jobs are done, the monitor job kicks off the FI job.
3) When the FI job is done, the monitor job returns the results.\

that seems like it'd be a lot less work than modifying EE2 (which IMO is the "best" solution, but also the biggest lift)

MrCreosote: Also why is the header prefixed with DEADLOCK?

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,139 @@
...
+## Intro
+Sometimes a calculation requires too many resources from one node (walltime, memory, disk), so the calculation gets spread across multiple machines.
+The final step of the app that uses KBParallels is to create a report. This step may use results from all of the computed jobs to create the final report.
+In order to do this, some apps use a mechanism called KBParallel. The apps are listed at [DATAUP-111](https://kbase-jira.atlassian.net/browse/DATAUP-111)
+
+## The current implementation of Batch Analysis in [kb_BatchApp](https://github.com/kbaseapps/kb_BatchApp) at KBase has the following issues:
+
+* Current UI is not adequate: Users shouldn’t have to code in order to run batch analysis. Also it’s difficult to do so, even for those familiar with KBase code (have to find object names)
+* Dependency on [KBParallel](https://github.com/kbaseapps/KBParallel): any changes to KBParallel could affect KB Batch and subsequently all other apps. 
+* Queue deadlocking: users have a max of 10 slots in the queue, with the current implementation one is taken up just to manage the jobs. Could lead to deadlock scenarios
```
MrCreosote, line 25: Can you quickly explain how this happens (e.g. 10 'management' jobs start before all the jobs for at least one of the management jobs finish, and so there are 10 management jobs running, and no 'real' jobs can start. Since the management jobs can't finish until their real jobs finish, the queue is deadlocked).

MrCreosote (changes requested):
This certainly seems good enough for a brain dump given that dealing with this problem is currently not in scope for the project. Just a few relatively minor changes and I think this is done

docs/adrs/004-SplitAndAggregate.md
bio-boris: done

docs/adrs/004-SplitAndAggregate.md
bio-boris: No it is not possible to kill them based on compute or disk, but it is possible to do it based on memory. One con would be that there would be a maximum amount of management jobs based on the number of cores, as most jobs require at least 1 core. It is possible to make the machine advertise a higher number of cores, but not sure how that would affect performance.

Because that solution only addresses the Deadlock, and not any UI issues. I  changed it a bit.

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,146 @@
...
+### Limit multiple instances of kbparallels
+
+* `+` Simplest solution, quickest turnaround, fixes deadlock issue
+* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs 
+
+### Increase number of slots or Seperate Queue for kbparallels apps without 10 job limit
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue
+* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
+* `-` A small amount of users can take over the entire system
+* `-` The calculations done by the apps will interfere with other apps and cause crashes/failures
```
MrCreosote, line 132: I think these two are true of we increase the number of slots, but not if KBP apps have a separate queue. Maybe those two solutions should be separate with individual pro/con lists?

MrCreosote (changes requested):

docs/adrs/004-SplitAndAggregate.md
bio-boris: @MrCreosote Which two?

bio-boris: changed

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -125,12 +125,21 @@ Still to be determined (not in scope of this ADR):
...
+### Increase number of slots per user > 10
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue for small numbers of jobs.
+* `-` Doesn't fix deadlock issue as the user can still submit more KBP jobs
 * `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
 * `-` A small amount of users can take over the entire system
 * `-` The calculations done by the apps will interfere with other apps and cause crashes/failures
 
+###  Seperate Queue for kbparallels apps that may or may not have its own limit to running jobs.
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue 
+* `+` Requires minimum changes to ee2 and condor if condor supports this feature
```
MrCreosote, line 137: What changes would it require?

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -125,12 +125,21 @@ Still to be determined (not in scope of this ADR):
...
+* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
+* `-` A small amount of users can take over the entire system unless  the new queue has its own limit to running jobs then it prevents users from taking over.
+* `-` The calculations done by the apps will interfere with other apps and cause crashes/failures
```
MrCreosote, line 140: This isn't true if it has its own queue, is it? Maybe I don't quite understand the separation between the queues.

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -125,12 +125,21 @@ Still to be determined (not in scope of this ADR):
...
+* `-` Doesn't fix deadlock issue as the user can still submit more KBP jobs
 * `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
 * `-` A small amount of users can take over the entire system
 * `-` The calculations done by the apps will interfere with other apps and cause crashes/failures
 
+###  Seperate Queue for kbparallels apps that may or may not have its own limit to running jobs.
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue 
+* `+` Requires minimum changes to ee2 and condor if condor supports this feature
+* `-` Addresses only the deadlocking issue, UI still broken for regular runs and batch runs
+* `-` A small amount of users can take over the entire system unless  the new queue has its own limit to running jobs then it prevents users from taking over.
```
MrCreosote, line 139: I don't think this is true, is it? They could dominate the KBP queue if there's no limit but the other queues should be immune. I think the answer here is there needs to be a job count limit on the KBP queue as well. Ideally we could have multiple jobs per processor as they're supposed to be light weight

docs/adrs/004-SplitAndAggregate.md
bio-boris: ee2 * Make a list of apps get submitted to that queue
condor * somehow make the 2nd queue a consumable resource

docs/adrs/004-SplitAndAggregate.md
bio-boris: fixed

bio-boris: changed

docs/adrs/004-SplitAndAggregate.md
bio-boris: Changed.
We are not working in the ideal situation, because right now the kbp jobs do computation as well. That would require rewriting all apps that use kbparallels that are doing computation which is not in the scope of this solution.

docs/adrs/004-SplitAndAggregate.md
MrCreosote: > Make a list of apps get submitted to that queue

can't you just specify that in the catalog?

> somehow make the 2nd queue a consumable resource

I don't understand what this means - why would this queue be any different than any other queue?

docs/adrs/004-SplitAndAggregate.md
MrCreosote: From the new text:

> The calculations done by the apps will interfere with other apps in the new queue still and cause crashes/failures since they are doing more than managing the jobs right now. This would be solved by not allowing multple instances of apps that use KBP, but then it would waste a lot of resources.

I guess I don't understand how this is different from any other queue.

MrCreosote:
maybe we should have a quick call and talk this out rather than going back and forth any more

bio-boris:
I thought of another way of doing it, so changed it around

docs/adrs/004-SplitAndAggregate.md
bio-boris: I guess I didn't mention that if we had a new queue, we would need to take the nodes from somewhere, and if we want to make better use of the nodes we take, we'd want to have more jobs running per KBP node.  Changed it around

docs/adrs/004-SplitAndAggregate.md
bio-boris: You could specify it in the catalog, and then make ee2 add a KBP_LIMIT if it has that QUEUE clientgroup, but if we want those KBP jobs to run in the NJS queue and limit the max of those jobs, EE2 would need to keep a list of the apps it needs to add the KBP_LIMIT to.

As for why making it a consumable resource, is because you don't want to have one user take over the entire KBP queue, since those jobs wouldn't take up slots anymore. If we limit the amount of jobs sent to the KBP queue, we might as well just not make a KBP queue, because we can limit KBP jobs without them needing their own queue.

MrCreosote: > As for why making it a consumable resource, is because you don't want to have one user take over the entire KBP queue, since those jobs wouldn't take up slots anymore. If we limit the amount of jobs sent to the KBP queue, we might as well just not make a KBP queue, because we can limit KBP jobs without them needing their own queue.

I'm still not quite grokking this - if we have a combined queue and a regular limit of 10 and a KBP limit of, say, 5, that means that a user running a lot of parallel jobs will be running at 50% capacity.

If instead we have a separate queue with a KBP limit of, say, 20, then the user can run at 100% capacity and run more KBP jobs at the same time. Theoretically we could also assign very low resources per KBP job (especially if we restructured the jobs to do most of the work on the regular queue).

bio-boris: * Modifying code inside of apps that use KBP is not in scope of this solution
What is a combined queue?

We have 60 cs nodes, which means we can have 60 jobs that take 60 nodes right now.
If we convert 20 nodes to KBP nodes and we had a limit of 20 jobs per user, then 1 user would submit 20 KBP jobs and no one else could run a KBP job. 20 KBP jobs would be set to running. Of those 20, only up to 10 could have a computation job running.  The other 10 would be doing nothing, as the users 10 slots are all taken up.

MrCreosote: Ok, gotcha, so it makes sense to make KBP_limit <= regular limit.

> If we convert 20 nodes to KBP nodes and we had a limit of 20 jobs per user, then 1 user would submit 20 KBP jobs and no one else could run a KBP job. 

This is where I'm wondering if we can break the 1 node : 1 job rule to allow more KBP jobs to run at the same time.

MrCreosote: > Modifying code inside of apps that use KBP is not in scope of this solution

Why not?

>  What is a combined queue?

I'm just talking about the case where KBP jobs and regular jobs run in the same queue.

bio-boris: Also FYI having more than 1 job do something like combine stuff with DFU or upload things on the same cs node might cause too much disk pressure on these older machines, so thats we changed to a max of 1 job per machine.

bio-boris: * This is where I'm wondering if we can break the 1 node : 1 job rule to allow more KBP jobs to run at the same time.
You sure can, but then you need to consider how many CPUS/MEMORY/DISK each KBP job is using and how many can fit in a node. If we do it wrong, then the jobs are all going to be affecting each other, causing more user tickets to come in complaining about their jobs mysteriously crashing. 

* Modifying code inside of apps that use KBP is not in scope of this solution
Well you have to cut off the solution at some point before you make a new one.
It seems very complex to come up with 7 plans of modifying KBP apps that make them not do any computation on the same machines and not affect other jobs, and weigh that against just ripping KBP out of those apps.

KBP and regular jobs already run in the same queue, see the proposal that allows the jobs to run on their own nodes and also fixes deadlocking *LIMIT KBP jobs to a maximum of 5 active jobs per user*

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+Between these there are several different ways to do what you want.
+
+There's a useful example here that shows the general workflow in the
+bindings: https://htcondor.readthedocs.io/en/latest/apis/python-bindings/tutorials/DAG-Creation-And-Submission.html#Describing-the-DAG-using-htcondor.dags
+
+## Consequences
+
+* We will have to roll out fixes in multiple stages
```
MrCreosote, line 106: Why is this?

bio-boris: Because Decision Outcome  has a solution with multiple stages of fixes.

MrCreosote: Oh, you're talking about the quick fix vs. long term fix thing. We don't *have* to do that though - we could just go straight for the long term fix.

bio-boris: True. Do you want me to remove this line?

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+* `+` List of apps is not frequently updated
+* `+` Apps do not need to be modified
+* `-` If a new app uses KBP and their app is not on the list, it won't be limited by the KBP_LIMIT unless the owner lets us know.
+* `-` If an existing app no longer uses KBP, their app is still limited unless the owner lets us know.
+* `-` Nodes will continue be taken up by jobs that do little computation as each job gets its own node.
+* `-` Users may not be able to effectively use up their 10 job spots 
+* `-` Capacity is still wasted, as some KBP jobs sit around waiting for other jobs to run
+
+###  LIMIT KBP jobs to a maximum of N<10 active jobs per user + Seperate queue for kbparallels apps 
+* `+` Same pros as above
```
MrCreosote, line 148: Pro: users will be able to effectively use their 10 job spots

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+#### Design of new verison
+* All apps must be modified to use the new KBP lightweight version, which will: 
+* Can either modify KBP, or create a new tool/package to use instead of KBP
+
+1) Launch a management job called the *Job Manager* that sits in the KBP Queue, alongside other KBP jobs. Other jobs are launched in the NJS queue.
+2) Launch the *Setup Job* which will
+ * Use the *User Parameters* and/or
+ * Optionally Download the Data from the *User Parameters*  to figure out *Job Manager* parameters
+ * Use the results of information gathered from the initial download and or *User Parameters*
+ * Generate final parameters to be sent to the *Job Manager* to launch *Fan Out* jobs
```
MrCreosote, line 168: The setup job could also just launch the jobs and return the job IDs to the job manager

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+3) The *Job Manager* now has enough parameters to setup *Fan Out* Jobs, and will send out jobs and wait for them (and possibly retry them upon failure)
+4) *Fan Out* jobs download data and perform calculations, save them back to the system, and return references to the saved objects
+5) The *Job Manager* optionally launches a *Group/Reduce* job based on User Parameters and or the results of *Fan Out* Jobs
+6) The *Group/Reduce* job downloads objects from the system, and creates a set or other grouping, and then saves the set back to the system
+7) The *Job Manager*  sends the results back to a *Report* Job, which downloads the final data and uploads a report based on that data
```
MrCreosote, line 173: I'm not sure why the report couldn't be created in the group/reduce job?

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+8) The *Job Manager* returns the reference to the results of the *Report Job*
+
+Pros/Cons
+* `-` Addresses the deadlocking issue, UI still broken for regular runs and batch runs if we re-use KBP
+* `-` Have to rewrite all apps that use KBP to use this new paradigm
```
MrCreosote, line 178: Note that this can be done on an as-needed basis; the apps don't all have to be converted at once

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+* `-` Will have to upload data and download data more times as no computation is done in the *Job Manager*
```
MrCreosote, line 179: I'm not sure I understand why this is is true. In current KBP, the data transfers are:
1) KBP node downloads data to calculate jobs - this will happen the same amount of times in LKBP (lightweight KBP), but in the setup job vs. the KBP job
2) PIP nodes download data, same as LKBP
3) KBP node downloads data to summarize / report - same for LKBP in the group/reduce if that makes the report as well

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+ 
+### Modify Apps to do only local submission by remove KBP, and moving the job 
+* `+` Simple solutions, quick turnarounds, fixes deadlock issue, fixes UI issues
+* `-` We have a limited number of larger resources machines
+* `-` Continued dependency on deprecated KBP tools
+* `-` App runs may take longer since fewer resources may be available to the app run
+
+### Deprecate kbparallels, and write new ee2 endpoints to entirely handle split and aggregate
+* `+` No longer uses an app, No longer uses a slot in the queue
+* `-` All jobs would have to change to stop using KBP in favor of these via the ee2 client,
```
MrCreosote, line 193: Note this can be done on an as-needed basis

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+* `+` Can manage jobs more closely: have tables, more control over lifecycle, resource requests, canceling subjobs
+* `-` Job monitoring thread would have to run in ee2
```
MrCreosote, line 195: I don't think this is true - EE2 would just have to check if all jobs are done when any PIP job is done. You don't need a thread for that.

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+* `-`  Requires new data structures and models in ee2 to make sure that the relationship is preserved between jobs and sub jobs, and to make sure deadlocking does not occur
+* `-` Requires storing the job outside of condor to prevent job submission, unless we can mark jobs as runnable or not via the HTCondor API*
```
MrCreosote, line 197: I don't understand this point - can you elaborate?

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,198 @@
...
+* `-` App runs may take longer since fewer resources may be available to the app run
+
+### Deprecate kbparallels, and write new ee2 endpoints to entirely handle split and aggregate
+* `+` No longer uses an app, No longer uses a slot in the queue
+* `-` All jobs would have to change to stop using KBP in favor of these via the ee2 client,
+* `+` Can manage jobs more closely: have tables, more control over lifecycle, resource requests, canceling subjobs
+* `-` Job monitoring thread would have to run in ee2
+* `-`  Requires new data structures and models in ee2 to make sure that the relationship is preserved between jobs and sub jobs, and to make sure deadlocking does not occur
+* `-` Requires storing the job outside of condor to prevent job submission, unless we can mark jobs as runnable or not via the HTCondor API* 
+* `-` The endpoint would need to know how to split up input files and aggregate them
```
MrCreosote, line 198: I don't this this is true - the FO/FI jobs would do that, right?

MrCreosote (changes requested):
Overall structure and contents looks good; I still have some questions / comments re the details

docs/adrs/004-SplitAndAggregate.md
MrCreosote: I guess so, it confused me anyway, and it's not strictly true.

docs/adrs/004-SplitAndAggregate.md
bio-boris: What if the report needs to have a value that is only know once the group/reduce job has completed?

docs/adrs/004-SplitAndAggregate.md
bio-boris: >  Will have to upload data and download data more times as no computation is done in the *Job Manager*
If the group reduce is done on objects, then no files have to be downloaded
if the group reduce is done on files that need to be downloaded from objects and then a new object is created you have to download these files to disk first.

docs/adrs/004-SplitAndAggregate.md
bio-boris: Ok, then I will remove it

docs/adrs/004-SplitAndAggregate.md
bio-boris: It won't know if it doesn't have the data to figure them out without downloading the file/s first

docs/adrs/004-SplitAndAggregate.md
bio-boris: Would you like me to add that?

docs/adrs/004-SplitAndAggregate.md
bio-boris: will add

docs/adrs/004-SplitAndAggregate.md
bio-boris: Yea this is not worded that well.......  Will elaborate

docs/adrs/004-SplitAndAggregate.md
MrCreosote: It has to do that anyway right? The difference we're talking about is whether
* The setup / FO job downloads data, figured out the parameters for the PIP jobs, and returns those parameters to EE2 to the LKBP job can use them to launch jobs and record the job IDs or
* The setup / FO job downloads data, figured out the parameters for the PIP jobs, launches the jobs, and returns the job IDS to EE2, which LKBP pulls from EE2 in order to trace the jobs.

docs/adrs/004-SplitAndAggregate.md
MrCreosote: But the group/reduce/FI job produces all the data / objects that are the end result of the job, so it has to know all the values before it ends execution, unless I'm missing something.

docs/adrs/004-SplitAndAggregate.md
MrCreosote: yes please

docs/adrs/004-SplitAndAggregate.md
MrCreosote: But that's true for both KBP and LKBP, unless I'm misunderstanding

docs/adrs/004-SplitAndAggregate.md
MrCreosote: yessir

docs/adrs/004-SplitAndAggregate.md
MrCreosote: That sounds right

docs/adrs/004-SplitAndAggregate.md
bio-boris: Sounds like we need to do another call

docs/adrs/004-SplitAndAggregate.md
bio-boris: Sounds like we should do another call

docs/adrs/004-SplitAndAggregate.md
MrCreosote: I thought this section was about https://github.com/kbase/execution_engine2/pull/422/files#diff-07a370a5473999326b83d1e3b2460686c97753837cfd37d1b78b12d809327ae5R63-R78

e.g. making EE2 do what the LKBP job would do, more or less, so you don't need the LKBP job (and probably makes UI stuff easier to handle if EE2 controls anything)

To me writing code in EE2 to figure out how to split up input for specific apps seems like a really bad idea; personally I wouldn't even have that as an option unless there's something I'm missing that makes it viable

bio-boris: Was this comment not about line 198, in the alternatives considered section?

MrCreosote: Yeah, I didn't realize that we were considering (I guess?) 2 different EE2 based solutions, one where EE2 just manages the jobs much like the LKBP job would, and one where EE2 actually does the FI and FO parts itself. If that's really what's going on here to my mind we could just not include the latter and I'd be fine with it, seems pretty crazypants to me.

Again, maybe I don't get it though. Let's talk in the call tomorrow

bio-boris: Ah, I think I see the ambiguity here. Basically instead of 
```
for item in data:
   params['job_input] = item
     ee2.run_job(params)
```
it would be
  ee2.run_fan_out_job(params, data)

```
def ee2_run_fan_out_job
    for item in data
    params[job_input] = item
   ee2.run_job()
```

docs/adrs/004-SplitAndAggregate.md
bio-boris: Ah I see what you mean, yes, we could send back job ids or parameters. Modified

docs/adrs/004-SplitAndAggregate.md
bio-boris: On second thought, if we send all the data from previous steps into the reduce job, it could do as you describe. However, if we don't have a reduce step, we'd still need a job to download the data and to create a report

bio-boris: Is there an assumption that we are always outputting a set from a KBP job?

MrCreosote: > However, if we don't have a reduce step, we'd still need a job to download the data and to create a report

To me both of those are just FI jobs, whether it's reduce, report, or reduce and report. For some jobs you might not need a FI step at all.

> Is there an assumption that we are always outputting a set and a report from a KBP job?

I don't think so.

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: We decided to delete this section in an OOB meeting

MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
MrCreosote: :+1:

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,189 @@
...
+* This will take significant time to further research and engineer the solutions
+
+Still to be determined (not in scope of this ADR): 
+* UI and how it relates to the bulk execution
+* XSV Analysis and how it relates to the bulk execution
+
+## Alternatives Considered
+
+* Ignore most issues and just make apps that run kbparallels limited to N instances of kbparallels per user to avoid deadlocks
+* Writing new ee2 endpoints to entirely handle batch execution and possibly use a DAG
```
MrCreosote, line 116: I think this line should be deleted, right?

docs/adrs/004-SplitAndAggregate.md
```diff
@@ -0,0 +1,189 @@
...
+ * Use the results of information gathered from the initial download and or *User Parameters*
+ * Generate final parameters to be sent to the *Job Manager* to launch *Fan Out* jobs, or directly launch *Fan Out* jobs and return job ids
+3) The *Job Manager* now has enough parameters to launch and/or monitor *Fan Out* Jobs, and monitor/manage the jobs (and possibly retry them upon failure)
+4) *Fan Out* jobs download data and perform calculations, save them back to the system, and return references to the saved objects
+5) The *Job Manager* launches one *FanIn* job based on User Parameters and or the results of *Fan Out* Jobs
+6) The *FanIn* (a.k.a Group/Reduce/Report) job downloads objects from the system, and creates a set or other grouping, and then saves the object(s) back to the system. Final data and report is uploaded back to the system
+8) The *Job Manager* returns the reference to the results of the *Report Job*
+
+Pros/Cons
+* `+` On an as needed basis, would have to rewrite apps that use KBP to use this new paradigm
```
MrCreosote, line 179: I think this is a con, but less of a con than having to rewrite all apps that use KBP at once

MrCreosote (changes requested):
Did a final read through and found a couple of minor things

MrCreosote (approved):
LGTM

PR #423, opened 2021-10-05
check_job_batch_bugaroo

bio-boris:
* Parameter got renamed. That means we missed the impl tests for this function.

### DIFF
--- a/test/tests_for_integration/api_to_db_test.py
+++ b/test/tests_for_integration/api_to_db_test.py
@@ -1362,8 +1362,32 @@ def test_run_job_batch(ee2_port, ws_controller, mongo_client):
 
         # Check to see check_job_batch has both the batch and child jobstates
         ret = ee2.check_job_batch(params={"job_id": batch_id})
-        assert "batch_jobstate" in ret
-        assert "child_jobstates" in ret
+        batch_jobstate = ret["batch_jobstate"]
+        child_jobstates = ret["child_jobstates"]
+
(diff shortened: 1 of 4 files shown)


MrCreosote (commented):
The code change looks fine, but I'd test it by just adding a check job call to the end of the run job batch tests in `api_to_db_test.py`. Needs less mocks that way, and I think the ones that are needed are already set up

test/tests_for_integration/api_to_db_test.py
```diff
@@ -1360,6 +1360,35 @@ def test_run_job_batch(ee2_port, ws_controller, mongo_client):
             sub_init, schedd_init, sub, schedd, txn, expected_sub_1, expected_sub_2
         )
 
+        # Check to see check_job_batch has both the batch and child jobstates
```
bio-boris, line 1363: In PyCharm, I can't even run these tests because This `assert_close_to_now` function fails because the test is so slow.. So I have to use `time PYTHONPATH=.:lib:test pytest -vv test/tests_for_integration/api_to_db_test.py::test_run_job_batch` which takes up to 50 seconds to run each time

What's interesting is that running that command is somehow triggering pytest in PyCharm to run, and I see a bunch of test failures appear in PyCharm every time too.

Between restarting docker and sometimes my entire computer each time I make changes to the api_to_db_test or if I do anything else in docker, it took 35 minutes to rewrite this simple test.

PR #425, opened 2021-11-04
Added indexes

bio-boris:

 

{"status": Status.queued.value, "queued": {"$lt": before_days}}

{"status": Status.created.value, "created": {"$lt": before_days}}

I don't think there's any indexes on status or the timestamp, so this is going to be a tablescan. I'd recommend making compound indices on state/queued and state/created

I'd just set them up in MongoUtil. That's more or less how I do it for all my servers, for example 
 

https://github.com/kbase/execution_engine2/blob/bc0644a997d6a39a4fbe62e3f7d4f93ec17d22f7/bin/PurgeBadJobs.py#L102

https://github.com/kbase/execution_engine2/blob/bc0644a997d6a39a4fbe62e3f7d4f93ec17d22f7/bin/PurgeBadJobs.py#L81

Hmm I wonder which type of indexes are best, and if I need a different index on 'created' because it actually needs to know if its a batch job or not.

### DIFF
--- a/lib/execution_engine2/db/models/models.py
+++ b/lib/execution_engine2/db/models/models.py
@@ -336,7 +336,11 @@ class Job(Document):
         default=False
     )  # Marked true when all retry steps have completed
 
-    meta = {"collection": "ee2_jobs"}
+    # See https://docs.mongoengine.org/guide/defining-documents.html#indexes
+    # Hmm, are these indexes need to be + or - ?
+    indexes = [("status", "-created"), ("status", "-queued")]
+
+    meta = {"collection": "ee2_jobs", "indexes": indexes}
 
     def save(self, *args, **kwargs):
         self.updated = time.time()


lib/execution_engine2/db/models/models.py
```diff
@@ -336,7 +336,11 @@ class Job(Document):
         default=False
     )  # Marked true when all retry steps have completed
 
-    meta = {"collection": "ee2_jobs"}
+    # See https://docs.mongoengine.org/guide/defining-documents.html#indexes
+    # Hmm, are these indexes need to be + or - ?
```
MrCreosote, line 340: Depends on if you want sorted results - if not, it doesn't really matter. With the status/queued index, if you do an equality query on status and a range query on queued you can sort descending without having to do an in-memory sort.

lib/execution_engine2/db/models/models.py
```diff
@@ -336,7 +336,11 @@ class Job(Document):
...
+    indexes = [("status", "batch_job"), ("status", "-queued")]
```
MrCreosote, line 341: Hmm - would it make sense to have an index with the username, e.g. `(user, status, -queued)`? Does EE2 not do any user specific queries?

bio-boris: Only for the job browser bff

```
        funcdef check_jobs_date_range_for_user(CheckJobsDateRangeParams params)
            returns (CheckJobsDateRangeResults) authentication required;
        funcdef check_jobs_date_range_for_all(CheckJobsDateRangeParams params)
            returns (CheckJobsDateRangeResults) authentication required;
```

lib/execution_engine2/db/models/models.py
MrCreosote: Ok, with the change now oldest first sorts can be done without an in-memory sort

lib/execution_engine2/db/models/models.py
```diff
@@ -336,7 +336,11 @@ class Job(Document):
         default=False
     )  # Marked true when all retry steps have completed
 
-    meta = {"collection": "ee2_jobs"}
+    # See https://docs.mongoengine.org/guide/defining-documents.html#indexes
+    # Hmm, are these indexes need to be + or - ?
+    indexes = [("status", "batch_job"), ("status", "queued")]
```
MrCreosote, line 341: What's the query for the status/batch_job index?

bio-boris: https://github.com/kbase/execution_engine2/blob/develop/bin/PurgeBadJobs.py#L103

lib/execution_engine2/db/models/models.py
MrCreosote: yeah I see that the original impetus was for the bad jobs reaper, which I assume doesn't care about the user name.

that being said, indexes targeted for those functions might not be a bad idea

MrCreosote (approved):
This looks fine to me assuming the indexes correctly handle the queries you expect

lib/execution_engine2/db/models/models.py
bio-boris: Here's another call that searches on clusterid instead of job id

https://github.com/kbase/execution_engine2/blob/1e268dfb1445e100da6fcb79219402953a32c0d1/lib/execution_engine2/sdk/EE2Status.py#L52-L53

bio-boris:
> This looks fine to me assuming the indexes correctly handle the queries you expect

Do we figure this out using .explain() ?

MrCreosote:
yep

MrCreosote:
Well, that and just thinking through how the queries will use the indexes and how selective the indexes will be for the query

PR #429, opened 2021-12-20
DATAUP-679: Add MongoUtil method to update a single job to queued

MrCreosote:

What's on the tin - add a method that updates a single job to the queued state. Next PR will update EE2RunJob to use it instead of the batched method.

# Testing Instructions

# Dev Checklist:

### DIFF
--- a/lib/execution_engine2/db/MongoUtil.py
+++ b/lib/execution_engine2/db/MongoUtil.py
@@ -272,7 +272,8 @@ class MongoUtil:
         return False
 
     def update_job_to_queued(
-            self, job_id: str, scheduler_id: str, scheduler_type: str = "condor") -> None:
+        self, job_id: str, scheduler_id: str, scheduler_type: str = "condor"
+    ) -> None:
         f"""
         * Updates a {Status.created.value} job to queued and sets scheduler state.
           Always sets scheduler state, but will only update to queued if the job is in the
@@ -292,13 +293,18 @@ class MongoUtil:
             # should we check that the job was updated and do something if it wasn't?
             ee2_jobs_col.update_one(
                 {"_id": ObjectId(job_id), "status": Status.created.value},
-                {"$set": {"status": Status.queued.value, "queued": queue_time_now}}
+                {"$set": {"status": Status.queued.value, "queued": queue_time_now}},
             )
             # originally had a single query, but seems safer to always record the scheduler
             # state no matter the state of the job
             ee2_jobs_col.update_one(
                 {"_id": ObjectId(job_id)},
-                {"$set": {"scheduler_id": scheduler_id, "scheduler_type": scheduler_type}}
+                {
+                    "$set": {
+                        "scheduler_id": scheduler_id,
+                        "scheduler_type": scheduler_type,
+                    }
+                },
             )
 
     def update_jobs_to_queued(
--- a/lib/execution_engine2/db/MongoUtil.py
+++ b/lib/execution_engine2/db/MongoUtil.py
@@ -271,6 +271,36 @@ class MongoUtil:
             return True
         return False
 
+    def update_job_to_queued(
+            self, job_id: str, scheduler_id: str, scheduler_type: str = "condor") -> None:
+        f"""
+        * Updates a {Status.created.value} job to queued and sets scheduler state.
+          Always sets scheduler state, but will only update to queued if the job is in the
+          {Status.created.value} state.
+        :param job_id: the ID of the job.
+        :param scheduler_id: the scheduler's job ID for the job.
+        :param scheduler_type: The scheduler this job was queued in, default condor
+        """
+        if not job_id or not scheduler_id or not scheduler_type:
+            raise ValueError("None of the 3 arguments can be falsy")
+        # could also test that the job ID is a valid job ID rather than having mongo throw an
+        # error
+        mongo_collection = self.config["mongo-jobs-collection"]
+        queue_time_now = time.time()
+        with self.pymongo_client(mongo_collection) as pymongo_client:
+            ee2_jobs_col = pymongo_client[self.mongo_database][mongo_collection]
+            # should we check that the job was updated and do something if it wasn't?
+            ee2_jobs_col.update_one(
+                {"_id": ObjectId(job_id), "status": Status.created.value},
+                {"$set": {"status": Status.queued.value, "queued": queue_time_now}}
+            )
+            # originally had a single query, but seems safer to always record the scheduler
+            # state no matter the state of the job
+            ee2_jobs_col.update_one(
+                {"_id": ObjectId(job_id)},
+                {"$set": {"scheduler_id": scheduler_id, "scheduler_type": scheduler_type}}
+            )
+
     def update_jobs_to_queued(
         self, job_id_pairs: List[JobIdPair], scheduler_type: str = "condor"
     ) -> None:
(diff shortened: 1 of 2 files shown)


bio-boris:
@MrCreosote You should probably use reuse https://github.com/kbase/execution_engine2/blob/develop/lib/execution_engine2/sdk/EE2Runjob.py#L989

bio-boris:
Oh wait, if the user batch cancels that won't work

MrCreosote:
That has a race condition so I think using `$set` is the right way to go. Also the new method duplicates the current batch behavior in MongoUtil except for it's targeted at a single job

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -271,6 +271,42 @@ def check_if_already_finished(job_status):
             return True
         return False
 
+    def update_job_to_queued(
```
bio-boris, line 274: This would need to return the status of the job in order to send the correct kafka message.
You could use the job object  if you wanted
J.queued = queue_time_now()
j.scheduler_id=scheduler_id
j.scheduler_type=condor

and then only set the status to queued if its "Created" or "Estimating" with the update_one statement

bio-boris: As the code stands it doesn't set the queued timestamp if the job was set to cancelled, even though it was queued and now has a job id.

MrCreosote: The batch method doesn't - it just does https://github.com/kbase/execution_engine2/blob/a7a2d8a5359aefed16cc8ded49573c76b4837515/lib/execution_engine2/sdk/EE2Runjob.py#L325-L335

My implementation in the next PR does the same thing

bio-boris: Right. It's kind of strange it it gets a scheduler id but doesn't have a queued timestamp, because how can you get a scheduler id without being queued?

bio-boris: But having a scheduler job queued is not the same as when the job record entered the "queued" state. So depending on how you interpret what that timestamp means, that behavior could be correct or incorrect.

bio-boris (approved):

lib/execution_engine2/db/MongoUtil.py
MrCreosote: > Right. It's kind of strange it it gets a scheduler id but doesn't have a queued timestamp, because how can you get a scheduler id without being queued?

That'll only happen if the job is set to running or canceled before Condor returns though. But if it is canceled, then you def want the scheduler info

MrCreosote: *if* the job is set to running prior to queued, we could add that other fix to `start_job` to set the queued time = running time, but from what you were saying it sounds like that should never happen under any reasonable circumstances

bio-boris: >That'll only happen if the job is set to running or canceled before Condor returns though. But if it is canceled, then you def want the scheduler info so you can cancel it scheduler-side

The condor submission doesn't change the status of the job or not. You can submit a job to condor that is labeled any state. It will start running and immediately die though, but will take up a slot in the queue unless it was already submitted in the queue, as condor_rm cannot remove a job if you don't know the scheduler id yet. 

Maybe we should change the state of the job before submitting it and never submit a job unless its in the queued state. If it's in the canceled state, don't submit a job.

MrCreosote: Mmm, then you could wind up with jobs in the queued state that aren't queued.

I was thinking maybe we should add a `submit` time but that's basically the same as `created` other than maybe a few ms

MrCreosote: Well, not for batch jobs. There could be a significant difference there. But I don't think the added complexity would actually help.

bio-boris: Should probably make a new ticket for this. Since it could take some time to submit a job to condor, this gives the opportunity for the fast cancel to at least prevent SOME of the jobs from actually being queued up in condor.

You wouldn't end up with jobs in the queued state that aren't queued because you would have to check the state multiple times.

Check state? Is it created? Then update to queued and add queued timestamp. Send kafka message.
Check state, is it queued? Then submit to condor, save scheduler id.

MrCreosote: > You can submit a job to condor that is labeled any state. It will start running and immediately die though

Let's say this happens:
* user submits job
* Condor submit starts
* User cancels job
* EE2 tries to update the job to the queued state, nothing happens

Now the job is queued on condor, and may stay queued for days or whatever, depending on how many jobs the user has scheduled. If the scheduler info is added to the job record, then the reaper thread could look for canceled jobs with jobs still in Condor and cancel them so they don't take a node (even for a short time).

MrCreosote: > Check state, is it queued? Then submit to condor...

If the cluster goes down here then you have a queued job record in Mongo but no job in Condor

MrCreosote: I kinda think we're overengineering here

bio-boris: I'm pretty sure that's even more reason to mark the job as queued before submitting it.

bio-boris: If mongo goes down, then the 14 day queued job reaper would clean it up.. Eventually.

MrCreosote: I don't understand that logic. If it's marked as created vs. queued, then that's accurate. There could even be code to recover the job on startup

MrCreosote: but again, I think we're solving problems that probably don't need to be solved

bio-boris: Was referring to https://github.com/kbase/execution_engine2/pull/429#discussion_r772081171

MrCreosote: Oh, for that case the job was never really in the queued state as far as the user is concerned, so I think jumping from created->canceled is fine. 

We might want to have a internal state for the Condor job (like a boolean saying whether it was canceled or not or something) that the reaper could use to clean up Condor jobs

bio-boris: The issue is that the thread is still going and submitting a bunch of cancelled jobs, because the condor job submission doesn't check the state of the job before submitting.

bio-boris: The boolean is a good idea, if no job scheduler id is saved or  condor is temporarily unavailable we could save that.

MrCreosote: > The issue is that the thread is still going and submitting a bunch of cancelled jobs, because the condor job submission doesn't check the state of the job before submitting. If we check the state before submitting, and its created or queued, then we are good to go.

Oh I see, just a short circuit. We could do that right now by just pulling each job before the submit and checking it's in the created state to avoid a cancel race condition.

Of course it's still possible for the cancel to happen between that check and Condor returning, so it doesn't completely fix the problem, but it reduces it, esp. for large batches.

On the other hand, we could just leave it to the reaper thread, depending on how often it runs - add a `canceled_on_scheduler` (or something like that) boolean to the job, and for any jobs that are canceled but that isn't true have the reaper kill the Condor job

bio-boris: yes exactly

PR #433, opened 2022-01-11
Don't toggle mongo collection instance variable

MrCreosote:

The current MongoUtil implementation has one `mongo_collection` instance
variable that can be set to either the jobs collection or the logs collection
depending on which method was called last. Methods that don't explicitly
choose the collection themselves will use the collection set by whichever
method was most recently called, which might be the wrong collection.

This commit makes two separate variables and ensures the right
collection is being used for each method. It also deletes a number of
unused methods.

# Testing Instructions

# Dev Checklist:

- [n/a] I have commented my code, particularly in hard-to-understand areas
- [?] My changes generate no new warnings - lots of warnings
- [n/a] I have added tests that prove my fix is effective or that my feature works - I could technically add a test, but it'd be very weird and wouldn't make a lot of sense IMO

- [x] [Version has been bumped](https://semver.org/) for each release
- [x] [Release notes](/RELEASE_NOTES.md) have been updated for each release (and during the merge of feature branches)

### DIFF
--- a/lib/execution_engine2/db/MongoUtil.py
+++ b/lib/execution_engine2/db/MongoUtil.py
@@ -33,7 +33,8 @@ class MongoUtil:
         self.mongo_pass = config["mongo-password"]
         self.retry_rewrites = parse_bool(config["mongo-retry-rewrites"])
         self.mongo_authmechanism = config["mongo-authmechanism"]
-        self.mongo_collection = None
+        self._col_jobs = config["mongo-jobs-collection"]
+        self._col_logs = config["mongo-logs-collection"]
         self._start_local_service()
         self.logger = logging.getLogger("ee2")
         self.pymongoc = self._get_pymongo_client()
@@ -155,16 +156,6 @@ class MongoUtil:
 
         return pymongo_client, mongoengine_client
 
-    @contextmanager
-    def pymongo_client(self, mongo_collection):
-        """
-        Instantiates a mongo client to be used as a context manager
-        Closes the connection at the end
-        :return:
-        """
-        self.mongo_collection = mongo_collection
-        yield self.pymongoc
-
     def get_workspace_jobs(self, workspace_id):
         with self.mongo_engine_connection():
             job_ids = [str(job.id) for job in Job.objects(wsid=workspace_id)]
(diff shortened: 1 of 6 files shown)


lib/execution_engine2/db/MongoUtil.py
```diff
@@ -155,41 +156,28 @@ def _get_collection(
...
 
-    @contextmanager
-    def pymongo_client(self, mongo_collection):
-        """
-        Instantiates a mongo client to be used as a context manager
-        Closes the connection at the end
-        :return:
-        """
-        self.mongo_collection = mongo_collection
-        yield self.pymongoc
```
MrCreosote, line 166: I've always been pretty confused about this point of this method, since it doesn't really do anything other than return the client. It doesn't actually close the client connection (which it shouldn't, so that's good) and doesn't do any context managing AFAICT. Setting the mongo collection instance variable here is what caused the problem so I decided to just delete the whole thing and simplify.

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -155,41 +156,28 @@ def _get_collection(
...
-
     def get_workspace_jobs(self, workspace_id):
         with self.mongo_engine_connection():
             job_ids = [str(job.id) for job in Job.objects(wsid=workspace_id)]
             return job_ids
 
     def get_job_log_pymongo(self, job_id: str = None):
```
MrCreosote, line 165: A lot of this is mostly whitespace changes - it'll be easier to read with whitespace off

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -290,88 +278,24 @@ def update_job_to_queued(
...
-                {"_id": ObjectId(job_id)},
-                {
-                    "$set": {
-                        "scheduler_id": scheduler_id,
-                        "scheduler_type": scheduler_type,
-                    }
-                },
-            )
-
-    def update_jobs_to_queued(
```
MrCreosote, line 314: Unused, replaced by the single job method above

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -542,26 +466,6 @@ def insert_jobs(self, jobs_to_insert: List[Job]) -> List[ObjectId]:
         inserted = Job.objects.insert(doc_or_docs=jobs_to_insert, load_bulk=False)
         return inserted
 
-    def insert_one(self, doc):
```
MrCreosote, line 545: All these low level DB methods are unused

lib/execution_engine2/db/MongoUtil.py
```diff
@@ -574,80 +478,15 @@ def _push_job_logs(self, log_lines: JobLog, job_id: str, record_count: int):
...
-                    projection=projection,
-                    batch_size=batch_size,
-                )
-            except Exception as e:
-                error_msg = "Connot query doc\n"
-                error_msg += "ERROR -- {}:\n{}".format(
-                    e, "".join(traceback.format_exception(None, e, e.__traceback__))
-                )
-                raise ValueError(error_msg)
+        job_col = self.pymongoc[self.mongo_database][self._col_logs]
```
MrCreosote, line 481: This method, if run when the old `mongo_collection` variable was set to the jobs collection, could add log information to the job record

MrCreosote: I also assume it'd fail with some sort of null pointer-ish error if it was called before one of the other methods that set `mongo_collection` had been called.

PR #464, opened 2023-02-15
Devops-960 Remove EE2 AUTOBUILD

bio-boris:
EE2 is automatically being built and over-writting ee2server which is bad because there are kb-sdk bugs there.


bio-boris:
Accidently pushed in the makefile :( without a PR

PR #482, opened 2024-08-02
D->M 0.0.14

bio-boris:
# 0.0.14
- Update clients to work with mongo7 by updating pymongo and mongoengine
-  Fix linting issues from flake8 and black
- Update outdated libraries that can no longer be pip installed
- The latest Dockerfile build contains conda which installs python 3.12.4 so these changes will pin python to 3.10 and save us headache in the future
- Speed up tests by moving docker build step out
- Create matrix of tests to run against mongo3.6 and mongo7

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+updates:
+
+  # Docker
+  - package-ecosystem: docker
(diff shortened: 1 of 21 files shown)


Xiangs18 (approved):
LGTM

